PR #3, opened 2015-05-15
Change syntax of licence

tcitworld:
I had the following error : 
License "AGPLv3" is not a valid SPDX license identifier, see http://www.spdx.org/licenses/ if you use an open license.


j0k3r:
:+1:

PR #10, opened 2015-08-21
Handle Safecurl

### LINKED ISSUE #7: Server Side Request Forgery (SSRF)
As [reported with wallabag](https://github.com/wallabag/wallabag/issues/1364#issuecomment-132507674), graby is vulnerable to SSRF. This means one can bypass restrictions for resources only accessible to localhost like  `http://127.0.0.1/server-status`,

j0k3r:
Should fix #7.

@freddyb Are you ok with that patch ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
     "require": {
         "php": ">=5.4",
         "guzzlehttp/guzzle": "^5.2.0",
+        "fin1te/safecurl": "~1.1",
         "simplepie/simplepie": "^1.3.1",
         "j0k3r/php-readability": "^1.0",
         "htmlawed/htmlawed": "^1.1.19",
--- a/composer.lock
+++ b/composer.lock
@@ -4,11 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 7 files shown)


freddyb:
I must admit that I don't understand enough of Graby to say with full confidence that this is correct.
If you could provide a smoke-test so a tester can throw URLs at Graby, like [DOMPurify does](https://cure53.de/purify), I would give it a test.

j0k3r:
All right, I've setup a dummy script that took the url given in parameter and give it to graby: http://kimtrip.net/graby/test.php?url=

```php
<?php

require 'vendor/autoload.php';

$graby = new \Graby\Graby(array('debug' => true));

try {
    $res = $graby->fetchContent($_GET['url']);
    var_dump($res);
} catch (\Exception $e) {
    var_dump($e);
}
```

freddyb:
Thanks! Taking a look.

freddyb:
I did some tests on SafeCurl in the meantime, looks good.

j0k3r:
So I guess that if you found SafeCurl good, my integration should be fine too.
Did you find time to take a look using the test url I provided?

freddyb:
Yeah, I did a cursory test to see this is showing similar results as the safecurl testbed at http://safecurl.fin1te.net/

Looks good.

j0k3r:
@freddyb great thanks !

PR #33, opened 2015-12-30
Refactored the ContentBuilder out of the ContentExtractor

bdunogier:
> Type: refactoring
> Backward compatible: yes

Moves the logic used in `ContentExtractor::buildSiteConfig()` to a new method in the `ContentBuilder`. I have left the HostFingerPrint part in the ContentExtractor, as I'm not 100% sure what it does (feedback welcome).

This change makes it possible to build site config objects from anywhere using the host name, as done by http://github.com/wallabag/wallabag in the site login feature PR.

### TODO
- [x] Fix tests, there are real failures (go over call logic in general)
- [x] Remove unrelated code about guzzle client

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -2,6 +2,7 @@
 
 namespace Graby;
 
+use Graby\SiteConfig\ConfigBuilder;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use GuzzleHttp\Client;
 use Readability\Readability;
@@ -29,11 +30,15 @@ class Graby
     private $httpClient = null;
     private $extractor = null;
 
+    /** @var \Graby\SiteConfig\ConfigBuilder */
+    private $configBuilder;
+
     /**
-     * @param array       $config
+     * @param array $config
      * @param Client|null $client Guzzle client
+     * @param \Graby\SiteConfig\ConfigBuilder $configBuilder
      */
-    public function __construct($config = array(), Client $client = null)
+    public function __construct($config = array(), Client $client = null, ConfigBuilder $configBuilder = null)
     {
         $resolver = new OptionsResolver();
         $resolver->setDefaults(array(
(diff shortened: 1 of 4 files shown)


src/Graby.php
```diff
@@ -74,11 +79,20 @@ public function __construct($config = array(), Client $client = null)
             $this->config['extractor'],
             $this->logger
         );
+
         $this->httpClient = new HttpClient(
-            $client ?: new Client(array('handler' => new SafeCurlHandler())),
```
bdunogier: This has nothing to do here, I'll remove it.

bdunogier:
Unsure about the message log test, will need to dig a bit deeper to fix the test.

bdunogier:
The tests pass locally, I'm a bit confused now.

bdunogier:
And the tests are passing.

j0k3r:
I'll have a look later this week :ok_hand:

j0k3r:
The HostFingerPrint check the given html and determine the website platform. In the default config, you have:

```php
'fingerprints' => array(
    '/\<meta\s*content=\"blogger\"\s*name=\"generator\"/i' => 'fingerprint.blogspot.com',
    '/\<meta\s*name=\"generator\"\s*content=\"Blogger\"/i' => 'fingerprint.blogspot.com',
    '/\<meta\s*name=\"generator\"\s*content=\"WordPress/i' => 'fingerprint.wordpress.com',
),
```

Which mean, we are looking for that html content (meta blabla) and if we find it we'll apply the value corresponding as the host and then apply the config file (for example) [`.wordpress.com.txt`](https://github.com/j0k3r/graby-site-config/blob/master/.wordpress.com.txt). This way, even if the url doesn't contains `wordpress.com` we can apply a wordpress site config on the blog.

Hope I'm clear :)

j0k3r:
Looks good to me, I'll tweaked few things after merge.
Thanks !

Can't wait to the future steps in Wallabag :)

PR #34, opened 2015-12-30
Use default guzzle cookies settings

bdunogier:
Changes the guzzle call to use the default cookie settings.
Also set the default guzzle client to accept cookies by default.

This allows customization of the guzzle client's cookies handling by providing a custom one with a CookieJar.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -75,9 +75,7 @@ class Graby
             $this->logger
         );
         $this->httpClient = new HttpClient(
-            $guzzleClient = $client ?: new Client(
-                array('handler' => new SafeCurlHandler(), 'defaults' => array('cookies' => true))
-            ),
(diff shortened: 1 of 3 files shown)


bdunogier:
Tests fixed (at least on HttpClient).

src/Graby.php
```diff
@@ -75,7 +75,9 @@ public function __construct($config = array(), Client $client = null)
             $this->logger
         );
         $this->httpClient = new HttpClient(
-            $client ?: new Client(array('handler' => new SafeCurlHandler())),
+            $guzzleClient = $client ?: new Client(
```
j0k3r: Why do you need to store the client in a variable?

bdunogier: Good question. Probably a leftover :-)

PR #35, opened 2015-12-30
Added login settings support

bdunogier:
> Required by wallabag's site login support.

Adds new settings that cover login to a site.

### Settings
- `requires_login`: boolean, indicates if the site requires login or not (default: false)
- `not_logged_in_xpath`: string, xpath query for an element in articles pages indicating that login is needed
- `login_uri`: string, the uri to which login data must be POSTed
- `login_username_field`: string, name of the username field in the form
- `login_password_field`: string, name of the password field in the form
- `login_extra_fields`: hash of form field => field value to POST with the login form

### DIFF
--- a/src/SiteConfig/SiteConfig.php
+++ b/src/SiteConfig/SiteConfig.php
@@ -87,11 +87,40 @@ class SiteConfig
     // the options below cannot be set in the config files which this class represents
     public $cache_key = null;
 
+    /**
+     * If fetching the site's content requires to authentify.
+     * @var bool
+     */
     public $requires_login = false;
+
+    /**
+     * XPath query to detect if login is requested in a page from the site.
+     * @var string
+     */
     public $not_logged_in_xpath = false;
+
+    /**
+     * Site's login form URI, if applicable.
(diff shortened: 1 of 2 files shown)


src/SiteConfig/SiteConfig.php
```diff
@@ -87,6 +87,13 @@ class SiteConfig
     // the options below cannot be set in the config files which this class represents
     public $cache_key = null;
 
+    public $requires_login = false;
```
j0k3r: Could you add a comment about how to use them? Like with a short example

j0k3r:
I've relaunched the build and it's OK now.

src/SiteConfig/SiteConfig.php
bdunogier: Added description of each setting.

j0k3r: Description looks good to me :+1:

bdunogier:
Is the coverage decrease a blocker ?

j0k3r:
Nope

PR #43, opened 2016-02-11
Fix tests without tidy

j0k3r:
It was (at least) the case for HHVM and for anyone who want to help without having tidy installed


j0k3r:
Shitty HHVM

PR #51, opened 2016-07-24
Handle space in url

j0k3r:
Handle spaces in URL to avoid unnecessary 404.

Related to: https://github.com/wallabag/wallabag/issues/2188

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -176,6 +176,7 @@ class Graby
         $response = $this->httpClient->fetch($url);
 
         $effectiveUrl = $response['effective_url'];
+        $effectiveUrl = str_replace(' ', '%20', $effectiveUrl);
         if (!$this->isUrlAllowed($effectiveUrl)) {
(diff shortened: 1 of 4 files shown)


src/Graby.php
```diff
@@ -176,6 +176,7 @@ private function doFetchContent($url)
         $response = $this->httpClient->fetch($url);
 
         $effectiveUrl = $response['effective_url'];
+        $effectiveUrl = str_replace(' ', '%20', $effectiveUrl);
```
tcitworld: May I ask why not directly use `url_encode() ?` :-)

j0k3r: Because it'll encode all the url and not only the part we want.
For an url like `http://lexpress.io/my awesome image.jpg` it'll generate `http%3A%2F%2Flexpress.io%2Fmy+awesome+image.jpg`.

tcitworld: Indeed, thanks !

PR #61, opened 2016-11-06
Add support of http_header property in site config

### LINKED ISSUE #49: add ability to send HTTP header like user-agent or referer
Hi, 

i would love to see graby being able to send additional http headers as configured in some ftr-site-config recipes.

Currently this is not supported:
https://github.com/j0k3r/graby/blob/026ffc79a6c014ab391e84ff1c85f7618bbc25e1/src/SiteConfig/SiteConfig.php#L39

There are some site configs wanting to send the user-agent: https://github.com/fivefilters/ftr-site-config/search?utf8=%E2%9C%93&q=user-agent

An example is also https://github.com/wallabag/wallabag/issues/2150 where the website thinks we are an internet explorer and we get redirected to hell.

Kdecherf:
This PR adds the ability to graby to use user-agent provided in site configs instead of `default_ua` and `user_agents` from `HttpClient`.

http_header is extracted from site config and provided to `HttpClient#fetch()` and `HttpClient#getUserAgent()` to extract the user-agent, if any. In case of missing value we fall back to the original behavior: check host from `user_agents` or use `default_ua`.

There are some issues however, at this time:
* ~~`default_ua` is not used anymore as long as a `http_header(user-agent)` is currently provided in `global.txt`. A solution could be to ignore the user-agent provided by global, any thoughts?~~
* `HttpClient` does not have access to SiteConfig, some tests began to fail because of the deletion of default `user_agents` array (e9f2981e573). I fixed them by providing a custom array with `user_agents` to the `HttpClient` instance in the test. Not sure if it's okay for you.

Fixes #49

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -52,19 +52,7 @@ class HttpClient
             // to see if returned content type matches $headerOnlyTypes.
             'header_only_clues' => array('pdf', 'mp3', 'zip', 'exe', 'gif', 'gzip', 'gz', 'jpeg', 'jpg', 'mpg', 'mpeg', 'png', 'ppt', 'mov'),
             // User Agent strings - mapping domain names
-            'user_agents' => array(
-                'lifehacker.com' => 'PHP/5.2',
-                'gawker.com' => 'PHP/5.2',
-                'deadspin.com' => 'PHP/5.2',
-                'kotaku.com' => 'PHP/5.2',
-                'jezebel.com' => 'PHP/5.2',
-                'io9.com' => 'PHP/5.2',
-                'jalopnik.com' => 'PHP/5.2',
-                'gizmodo.com' => 'PHP/5.2',
-                '.wikipedia.org' => 'Mozilla/5.2',
-                '.fok.nl' => 'Googlebot/2.1',
-                'getpocket.com' => 'PHP/5.2',
-            ),
+            'user_agents' => array(),
             // AJAX triggers to search for.
             // for AJAX sites, e.g. Blogger with its dynamic views templates.
             'ajax_triggers' => array(
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -360,10 +360,12 @@ class HttpClient
 
         foreach ($try as $h) {
             if (isset($this->config['user_agents'][$h])) {
+                $this->logger->log('debug', 'Found user-agent "{user-agent}" for url "{url}" from config', array('user-agent' => $this->config['user_agents'][$h], 'url' => $url));
                 return $this->config['user_agents'][$h];
             }
         }
 
+        $this->logger->log('debug', 'Use default user-agent "{user-agent}" for url "{url}"', array('user-agent' => $ua, 'url' => $url));
         return $ua;
     }
 
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -101,10 +101,11 @@ class HttpClient
      *
      * @param string $url
      * @param bool   $skipTypeVerification Avoid mime detection which means, force GET instead of potential HEAD
+     * @param array  $httpHeader Custom HTTP Headers from SiteConfig
      *
      * @return array With keys effective_url, body & headers
      */
-    public function fetch($url, $skipTypeVerification = false)
+    public function fetch($url, $skipTypeVerification = false, $httpHeader = array())
     {
         if (false === $this->checkNumberRedirects($url)) {
             return $this->sendResults(array(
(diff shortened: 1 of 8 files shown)


Kdecherf:
Well, nevermind for the first issue. It was a non-tracked manual change in `vendor/j0ker/graby-site-config/global.txt` on my side. Anyway, I don't know if you are okay with this behavior: `user_agents` and `default_ua` are completely ignored if global.txt provides a `http_header(user-agent)`, which is not the case actually (for now).

src/Extractor/HttpClient.php
```diff
@@ -100,11 +88,12 @@ public function setLogger(LoggerInterface $logger)
      *     - content type header.
      *
      * @param string $url
+     * @param array  $http_header Custom HTTP Headers from SiteConfig
      * @param bool   $skipTypeVerification Avoid mime detection which means, force GET instead of potential HEAD
      *
      * @return array With keys effective_url, body & headers
      */
-    public function fetch($url, $skipTypeVerification = false)
+    public function fetch($url, $http_header = array(), $skipTypeVerification = false)
```
j0k3r: Can you use `httpHeader` instead?
All variables are lowerCamelCase.

src/SiteConfig/ConfigBuilder.php
```diff
@@ -341,6 +345,8 @@ public function parseLines(array $lines)
             } elseif ((substr($command, -1) == ')') && preg_match('!^([a-z0-9_]+)\((.*?)\)$!i', $command, $match) && $match[1] == 'replace_string') {
                 array_push($config->find_string, $match[2]);
                 array_push($config->replace_string, $val);
+            } elseif ((substr($command, -1) == ')') && preg_match('!^([a-z0-9_]+)\(([a-z0-9_-]+)\)$!i', $command, $match) && $match[1] == 'http_header' && in_array($match[2], array('user-agent'))) {
```
j0k3r: Instead of `in_array($match[2], array('user-agent'))` can you use a simple `'user-agent' === $match[2]`?

Kdecherf: I used `in_array()` to make addition of new headers easier. But I can replace it with a strict check if you prefer.

src/Extractor/HttpClient.php
```diff
@@ -52,19 +52,7 @@ public function __construct(Client $client, $config = array(), LoggerInterface $
...
-                'deadspin.com' => 'PHP/5.2',
-                'kotaku.com' => 'PHP/5.2',
-                'jezebel.com' => 'PHP/5.2',
-                'io9.com' => 'PHP/5.2',
-                'jalopnik.com' => 'PHP/5.2',
-                'gizmodo.com' => 'PHP/5.2',
-                '.wikipedia.org' => 'Mozilla/5.2',
-                '.fok.nl' => 'Googlebot/2.1',
-                'getpocket.com' => 'PHP/5.2',
-            ),
```
j0k3r: If we remove that, we should be sure that these values are in the config file for each domain.

src/Extractor/HttpClient.php
```diff
@@ -100,11 +88,12 @@ public function setLogger(LoggerInterface $logger)
      *     - content type header.
      *
      * @param string $url
+     * @param array  $httpHeader Custom HTTP Headers from SiteConfig
      * @param bool   $skipTypeVerification Avoid mime detection which means, force GET instead of potential HEAD
      *
      * @return array With keys effective_url, body & headers
      */
-    public function fetch($url, $skipTypeVerification = false)
+    public function fetch($url, $httpHeader = array(), $skipTypeVerification = false)
```
j0k3r: Could you move that new parameter as third one? If we inject it in the second arg, it'll require to release a new major version because the signature of this public method will change. I know, it'll require to put the `false` argument for `$skipTypeVerification` every where you'll have to inject `$httpHeader` but it'll avoid BC

src/Extractor/HttpClient.php
```diff
@@ -313,12 +302,19 @@ private function possibleUnsupportedType($url)
...
+     * @param array  $httpHeader Custom HTTP Headers from SiteConfig
      *
      * @return string
      */
-    private function getUserAgent($url)
+    private function getUserAgent($url, $httpHeader = array())
     {
         $ua = $this->config['ua_browser'];
+
+        if (array_key_exists('user-agent', $httpHeader) && !empty($httpHeader['user-agent'])) {
```
j0k3r: You don't need the `array_key_exists()`, `empty()` already perform the `isset()`.

src/Extractor/HttpClient.php
j0k3r: Ok, skip it, I've checked, they are all already defined in the site config 👍

j0k3r:
> I fixed them by providing a custom array with `user_agents` to the `HttpClient` instance in the test. Not sure if it's okay for you.

Yeah it's ok, you should have an other option to inject it in the `fetch()` but what you've done allow the fallback to the old behavior to still be tested

PR #80, opened 2017-02-24
Use wallabag/tcpdf

j0k3r:
tecnickcom/tcpdf is not maintain anymore.
Developer has moved to a new library which is far from being stable atm.
Instead of keeping an unmaintain deps, the wallabag/core team forked the original repo to submit patch.

See https://github.com/wallabag/wallabag/issues/2930

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "symfony/options-resolver": "~2.6|~3.0",
         "monolog/monolog": "^1.13.1",
         "smalot/pdfparser": "~0.9.24",
+        "wallabag/tcpdf": "^6.2",
         "true/punycode": "~2.0"
     },
     "require-dev": {


jtojnar:
What is tcpdf for? Is it used somewhere?

tcitworld:
To fetch pdfs

j0k3r:
To _read_ pdfs

jtojnar:
Isn’t PdfParser enough? I cannot find where is TCPDF used.

jtojnar:
Oh, I see. It is replacing the dependency of the PdfParser, that’s why I could not grep it.

jtojnar:
Well, I decided to [extract the parser](https://packagist.org/packages/fossar/tcpdf-parser) but it does not seem to work anyway.

j0k3r:
What are you trying to achieve?

jtojnar:
I migrated selfoss to use Graby but I do not want to include the whole tcpdf since it is massive.

For that reason, I decided to strip the library to only include the parser but I found it does not work anyway – TCPDF_PARSER ERROR: Empty PDF data.

PR #82, opened 2017-03-07
Fixed xss-filter part which was useless

nicosomb:
Graby has a xss-filter parameter in the configuration but it was useless because the sanitized text was not re-used.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -147,7 +147,7 @@ class Graby
         // filter xss?
         if ($this->config['xss_filter']) {
             $this->logger->log('debug', 'Filtering HTML to remove XSS');
-            $html = htmLawed($html, array(
+            $infos['html'] = htmLawed($html, array(
                 'safe' => 1,
                 'deny_attribute' => 'style',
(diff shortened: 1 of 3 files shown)


src/Graby.php
```diff
@@ -155,6 +155,8 @@ public function fetchContent($url)
             ));
         }
 
+        $infos['html'] = $html;
```
j0k3r: I would rather override `$infos['html']`.

```php
$infos = $this->doFetchContent($url);

// filter xss?
if ($this->config['xss_filter']) {
    $this->logger->log('debug', 'Filtering HTML to remove XSS');

    $infos['html'] = htmLawed($infos['html'], array(
        'safe' => 1,
        'deny_attribute' => 'style',
        'comment' => 1,
        'cdata' => 1,
    ));
}

// generate summary
$infos['summary'] = $this->getExcerpt($infos['html']);
```

nicosomb:
Ready to be squashed.

j0k3r:
Squash allowed 🚀

PR #87, opened 2017-03-18
Use parseContent method for parsing PDFs

jtojnar:
Since the response body is available, there is no need to re-fetch the resource again. Additionally, the PDF parser [does not handle errors when fetching files](https://github.com/smalot/pdfparser/blob/c20bfd40e3d7f40ff4083de5971b2a9e7a839dd5/src/Smalot/PdfParser/Parser.php#L72), so this also improves error handling. This is the brokenness I mentioned [here](https://github.com/j0k3r/graby/pull/80#issuecomment-287549743), the server returned 403 but PdfParser complained about being unable to parse.

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -42,7 +42,6 @@ class HttpClient
             // HTTP responses which match these content types will
             // be returned without body.
             'header_only_types' => array(
-                'application/pdf',
                 'image',
                 'audio',
                 'video',
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -47,9 +47,9 @@ class HttpClient
                 'video',
             ),
             // URLs ending with one of these extensions will
-            // prompt Humble HTTP Agent to send a HEAD request first
+            // prompt client to send a HEAD request first
             // to see if returned content type matches $headerOnlyTypes.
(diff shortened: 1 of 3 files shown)


jtojnar:
I see the re-fetching should not be a problem, since in most cases HEAD method is used:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use Graby\Graby;

$g = new Graby(['debug' => true]);

// This works, probably because it is not recognized as a PDF due to redirects
$response = $g->fetchContent('http://arxiv.org/pdf/1703.05607');
var_dump($response);

// While this fails
$response = $g->fetchContent('https://arxiv.org/pdf/1703.05607.pdf');
```

Making PdfParser throw Exceptions would fix the error reporting problem. However, since PdfParser uses `file_get_contents`, it is not able to use the client configuration (user-agent) passed to Graby and there might be other problems such as broken redirect following – downloading the data in Graby is probably a better way. Would removing `application/pdf` from the [`header_only_types`](https://github.com/j0k3r/graby/blob/b747c461959e24a9f82d5aa3583fdce8036cd17e/src/Extractor/HttpClient.php#L44) be an acceptable solution?

j0k3r:
Looks like a good idea.
Also it'll save one http request.
I've fixed one thing about extension and also tests.

Issue #85, opened 2017-03-14
Get language from headers


ngosang:
Currently you are getting the language information from the "lang" tag, but many sites doesn't have this tag and uses HTTP headers instead.
My proposal is to use the "lang" tag and if it's not available get the language from the header "Content-Language".
Example URl without tag but with header:
http://www.investopedia.com/university/introduction-stock-trader-types/pivot-traders.asp

Already reported in wallabag but they send me here => https://github.com/wallabag/wallabag/issues/2978

j0k3r:
Might be a good idea :+1: 
For the record:

```
HTTP/1.1 200 OK
Date: Tue, 14 Mar 2017 09:41:21 GMT
Server: Apache
Expires: Sun, 19 Nov 1978 05:00:00 GMT
Last-Modified: Tue, 14 Mar 2017 09:41:21 +0000
Cache-Control: no-cache, must-revalidate, post-check=0, pre-check=0
Content-Language: en
Content-Type: text/html; charset=utf-8
Vary: Accept-Encoding
ETag: W/"1489484481-gzip"
X-Varnish: 59793755 59693389
Age: 5488
Via: 1.1 varnish-v4
X-Dev: true
X-Cookie: 
X-UA-Device: desktop
X-VH: 1
X-SEM: false
X-Cache: HIT
X-Cache-Hits: 7
Transfer-Encoding: chunked
Connection: keep-alive
```

nicosomb:
Useful for me because I want to know if the http headers contain "X-Frame-Options" value.

PR #97, opened 2017-04-29
Return all headers

j0k3r:
Follow up https://github.com/j0k3r/graby/pull/96

Return all headers from the response to the client.
This will allow more stuff to me done on the client side depending on the headers.

Also, it'll allow to handle https://github.com/j0k3r/graby/issues/85


nicosomb:
Thank you @j0k3r

PR #98, opened 2017-04-29
Retrieve language from header

j0k3r:
In case if we failed to retrieve it from the content.

Fix https://github.com/j0k3r/graby/issues/85


ngosang:
@j0k3r thank you.

PR #99, opened 2017-05-05
Switch to MIT

j0k3r:
Following https://github.com/wallabag/wallabag/issues/179#issuecomment-299417553 I'm switching the graby license from AGPL to MIT


tcitworld:
The number of characters diff says a lot. :sneezing_face:

PR #102, opened 2017-05-15
Implemented extra config builder extension point

bdunogier:
Third parties can register extra configuration, given a prefix and an `ExtraConfigBuilder`.
All commands starting with the given prefix will be sent to the matching builder, that will return a custom `ExtraSiteConfig` object.

Example:

```
$configBuilder->addExtraConfigBuilder('paywall', new PaywallExtraConfigBuilder());
```

Implementors are responsible for defining their own ExtraSiteConfig object, as well as their own
`ExtraConfigBuilder`.

### DIFF
--- a/src/SiteConfig/ConfigBuilder.php
+++ b/src/SiteConfig/ConfigBuilder.php
@@ -16,6 +16,7 @@ class ConfigBuilder
 
     /**
      * Third party config builders, indexed by their prefix.
+     *
      * @var \Graby\SiteConfig\ExtraConfigBuilder[]
      */
     private $extraConfigBuilders = [];
@@ -351,7 +352,7 @@ class ConfigBuilder
             }
 
             foreach (array_keys($this->extraConfigBuilders) as $prefix) {
-                if (strpos($command, $prefix . "_") === false) {
+                if (strpos($command, $prefix.'_') === false) {
                     continue;
                 }
                 $command = substr($command, strlen($prefix) + 1);
@@ -360,7 +361,7 @@ class ConfigBuilder
                 } elseif (!is_array($extraConfigItems[$prefix][$command])) {
                     $extraConfigItems[$prefix][$command] = [
                         $extraConfigItems[$prefix][$command],
-                        $val
+                        $val,
                     ];
                 } else {
                     $extraConfigItems[$prefix][$command][] = $val;
--- a/src/SiteConfig/ConfigBuilder.php
+++ b/src/SiteConfig/ConfigBuilder.php
@@ -15,6 +15,12 @@ class ConfigBuilder
     private $cache = array();
 
     /**
+     * Third party config builders, indexed by their prefix.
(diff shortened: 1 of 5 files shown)


src/SiteConfig/ConfigBuilder.php
```diff
@@ -332,6 +350,23 @@ public function parseLines(array $lines)
                 continue;
             }
 
+            foreach (array_keys($this->extraConfigBuilders) as $prefix) {
+                if (strpos($command, $prefix . "_") === 0) {
```
j0k3r: You can switch the condition and made an early `continue`

src/SiteConfig/ExtraConfigBuilder.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
```
j0k3r: Hum

src/SiteConfig/ExtraConfigBuilder.php
```diff
@@ -0,0 +1,20 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace Graby\SiteConfig;
+
+
```
j0k3r: You can remove that empty line

j0k3r (commented):
Could you add some tests with a real `ExtraConfigBuilder` class which returns a real `SiteExtraConfig`?

j0k3r:
Also, don't forget to run `php vendor/bin/php-cs-fixer fix` too.

src/SiteConfig/ExtraConfigBuilder.php
bdunogier: rofl, sorry :)

bdunogier:
I didn't really clean it up before sending it. The idea was to let you see the big picture.
I'll take care of it.

bdunogier:
> Could you add some tests with a real ExtraConfigBuilder class which returns a real SiteExtraConfig?

I will. Review remarks fixed in the meanwhile.

src/SiteConfig/ConfigBuilder.php
```diff
@@ -332,6 +350,24 @@ public function parseLines(array $lines)
...
+                    $extraConfigItems[$prefix][$command] = $val;
+                } elseif (!is_array($extraConfigItems[$prefix][$command])) {
+                    $extraConfigItems[$prefix][$command] = [
+                        $extraConfigItems[$prefix][$command],
+                        $val
+                    ];
+                } else {
+                    $extraConfigItems[$prefix][$command][] = $val;
+                }
+                continue 2;
```
bdunogier: Nope. A single `continue` would continue from `foreach (array_keys($this->extraConfigBuilders) as $prefix)`. As we have found an extra config matching the prefix, we want to continue from the next line (`foreach ($lines as $line)`).

PR #104, opened 2017-05-20
Update README.md

bbapp:


j0k3r:
Why?

bbapp:
I just thought it was worth noting that Graby is a better version of PHP Readability since the two libraries appear to do the same thing. However, Graby is based on Full-Text RSS v3.3 which would suggest a product specifically meant for RSS feeds. I was confused when I discovered this repo and I just thought it would be helpful to add a few words to the readme.txt to add clarity.

I honestly thought this would be self-explanatory.

j0k3r:
Thanks for the explanation, I've rewritten the top of the readme.
Is it more relevant to you?

bbapp:
Yeah, you changes makes things much more clear. Also, I noticed the array tag with missing from the example PHP code, so I submitted the changes.

PR #105, opened 2017-05-24
Ability to clean given HTML

j0k3r:
Add ability to clean a given HTML content using same method as when fetching content from an url.
For example it'll allow wallabag to save content without fetching url (when using from a browser extension for example).

Code as been a refactored to move clean logic into a dedicated method.

Also, the method `handleMimeAction` is now more clear and less complex.

@tcitworld @nicosomb what crazy html content should I check for that particular case?
I've added some obvious xss using `<script>` tag but I think we can found more malicious HTML :skull_and_crossbones:

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -165,6 +165,11 @@ class Graby
             $contentBlock = $this->extractor->getContent();
         }
 
+        // in case of extractor failed
+        if (null === $contentBlock) {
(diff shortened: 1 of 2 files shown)


nicosomb:
> what crazy html content should I check for that particular case?

No idea for that. Maybe you can have a look to other libraries which handle this behavior.

PR #110, opened 2017-06-05
Fix utf8 encoding for plain text page

aaa2000:
when the body is serialized, the data could not be encoded
because it contains invalid UTF8 characters.

@see https://github.com/wallabag/wallabag/issues/3154


j0k3r:
Looks good.
You just need to fix the CS issue.

aaa2000:
Done

j0k3r:
Thanks!

PR #129, opened 2017-11-03
Agnostic Guzzle

j0k3r:
- Convert hard coded Guzzle 5 test
- Update timeout in README, the full config might be at the end of the README
- Be Guzzle version agnostic
- Add a Travis build with Guzzle 6 (since Guzzle 5 is by default)

### DIFF
--- a/tests/Extractor/HttpClientTest.php
+++ b/tests/Extractor/HttpClientTest.php
@@ -5,7 +5,6 @@ namespace Tests\Graby\Extractor;
 use Graby\Extractor\HttpClient;
 use GuzzleHttp\Client as GuzzleClient;
 use GuzzleHttp\Psr7\Response;
-use Http\Adapter\Guzzle5\Client as GuzzleAdapter;
 use Http\Mock\Client as HttpMockClient;
 use Monolog\Handler\TestHandler;
 use Monolog\Logger;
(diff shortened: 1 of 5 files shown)


tests/Extractor/HttpClientTest.php
```diff
@@ -290,8 +288,25 @@ public function testTimeout()
         $handler = new TestHandler();
         $logger->pushHandler($handler);
 
-        $guzzle = new GuzzleClient(['defaults' => ['timeout' => 2]]);
-        $adapter = new GuzzleAdapter($guzzle);
+        // find with adapter is installed
+        if (class_exists('Http\Adapter\Guzzle6\Client')) {
```
aaa2000: it's almost like the discovery of php-http http://docs.php-http.org/en/latest/discovery.html but if you want configure the timeout, I think there is no other way

tests/Extractor/HttpClientTest.php
```diff
@@ -290,8 +288,25 @@ public function testTimeout()
...
+        } elseif (class_exists('Http\Client\Curl\Client')) {
+            $adapter = new \Http\Client\Curl\Client(
+                \Http\Discovery\MessageFactoryDiscovery::find(),
+                \Http\Discovery\StreamFactoryDiscovery::find(),
+                [
+                    CURLOPT_TIMEOUT => 2,
+                ]
+            );
+        } else {
+            $this->markTestSkipped('No Guzzle adapter defined ?');
```
aaa2000: Else instead of skip, you can do `$adapter = Http\Discovery\HttpClientDiscovery::find();`. So if it need to test with other adapter, it is necessary only to modify travis fail.  Only valid if the 2s of timeout is not a requirements by the tests

tests/Extractor/HttpClientTest.php
```diff
@@ -303,7 +318,8 @@ public function testTimeout()
 
         $this->assertSame('Request throw exception (with no response): {error_message}', $records[3]['message']);
         // cURL error 28 is: CURLE_OPERATION_TIMEDOUT
-        $this->assertContains('cURL error 28', $records[3]['formatted']);
+        // "cURL error 28: Connection timed out after"
+        $this->assertContains('Connection timed out after', $records[3]['formatted']);
```
aaa2000: The timeout is necessary and each php-http client has diferent messages  https://github.com/php-http/httplug/issues/118 so the tests will not pass with a dfferent $adapter that guzzle and curl so my previous is obsolete

j0k3r: And if we really want to add a new adapter, we'll update the assertion accordingly

PR #130, opened 2017-11-06
Require tidy extension for tests

aaa2000:
Fix #126

PR #132, opened 2017-11-10
Add a small message about Tidy

j0k3r:
Following discussion here: https://github.com/j0k3r/graby/pull/130
Poke @Kdecherf

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,12 @@ That's why I made this fork:
 
 ## How to use it
 
+### Requirements
+
+By default, this lib will use the [Tidy extension](https://github.com/htacg/tidy-html5) if it's available. Tidy is only used to cleanup the given HTML and avoid problems with bad HTML structure, etc .. It'll be suggested by Composer.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,12 @@ That's why I made this fork:
 
 ## How to use it
 
+### Requirements
+
+By default, this lib will use the [Tidy extension](https://github.com/htacg/tidy-html5) if it's available. Tidy is only used to cleanup the given HTML and avoid problems with bad HTML structure, etc .. It'll be suggested by Composer.
+
```
aaa2000: Should not it be added  in `composer.json`  like `php-readability`  ?

```
"suggest": {
        "ext-tidy": "Used to clean up given HTML and to avoid problems with bad HTML structure."
}
```

j0k3r: Since graby require php-readability, the message already exists when you install graby:

```
j0k3r/php-readability suggests installing ext-tidy (Used to clean up given HTML and to avoid problems with bad HTML structure.)
```

aaa2000: ok

Issue #138, opened 2017-12-06
Switch to newer HTMLawed


jtojnar:
`htmlawed/htmlawed` composer package is not compatible with PHP 7.2 and the package maintainer is not responsive. Since PHP 7.2 was released, I [forked the package](https://packagist.org/packages/fossar/htmlawed). I can make you a repo co-owner, if you wish.

j0k3r:
Good idea. Ok for the co-owner (even if I found the code very ugly and almost unmaintainable ...).
Feel free to open a PR to update composer to use your fork :+1: 
Also, maybe you should add 7.2 (and maybe allowed it to fail) in the .travis.yml file to test graby against that version of PHP

jtojnar:
The issue with HTMLawed is that it is not developed publicly, the GitHub repositories basically just copy the files from upstream and publish them on Composer.

I also added you as a maintainer, though not sure if I managed to set the permissions correctly.

j0k3r:
So you just updated the code from that private repository and didn't fix any issue on your own?
I thought you did all changes on your own in that PR https://github.com/kesar/HTMLawed/pull/16 :sweat_smile:

jtojnar:
We might try to convince upstream to release an official composer package. This is more of a mid-term measure.

jtojnar:
We might try to convince upstream to release an official composer package. This is more of a mid-term measure for PHP 7.2 compatibility.

j0k3r:
Posted http://www.bioinformatics.org/phplabware/forum/viewtopic.php?pid=904
We'll see ...

j0k3r:
http://www.bioinformatics.org/phplabware/forum/viewtopic.php?pid=907#p907 :disappointed:

aaa2000:
> I have looked into Packagist, and you are right that it appears unusable if source code is not on GitHub but on a self-hosted site.

it does not need to be hosted on github, it can be a VCS repository https://getcomposer.org/doc/05-repositories.md

j0k3r:
@aaa2000 you're right, maybe you can post a reply to the thread so it'll see that more people are engaged into that composer package

Zegnat:
Just ran into this and have written about [an alternative solution on the Bioinformatics forum](http://www.bioinformatics.org/phplabware/forum/viewtopic.php?id=325).

j0k3r:
I've chosen to use @jtojnar solution.
The one from Bioinformatics looks too complex.
Release in 1.12.1.

PR #139, opened 2017-12-08
Php 7.2

j0k3r:
Related https://github.com/j0k3r/graby/issues/138

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ php:
     - 5.6
     - 7.0
     - 7.1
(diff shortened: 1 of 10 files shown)


tcitworld:
>  Use namespaced PHPUnit classes 

This feels sooo good.

aaa2000:
Maybe, we could use htmlpurifier instead of htmLawed  ? http://htmlpurifier.org/comparison#Summary

j0k3r:
It could be an idea for the 2.0

PR #141, opened 2018-01-23
Empty nodes removed (related to wallabag #3455)

Simounet:
Just as I told you, this is the way I found to avoid empty node being fetched. It can fix issues like [this one](https://github.com/wallabag/wallabag/issues/3455). If you have a better idea, let me know.

Have a nice day.

### DIFF
--- a/tests/GrabyTest.php
+++ b/tests/GrabyTest.php
@@ -1436,4 +1436,23 @@ class GrabyTest extends TestCase
         $this->assertArrayHasKey('html', $res);
         $this->assertNotFalse(json_encode($res['html']), json_last_error_msg());
     }
+
+    public function testEmptyNodesRemoved()
+    {
+        $response = new Response(
+            200,
+            [
+                'content-type' => 'text/html',
+            ],
+            Stream::factory(file_get_contents(__DIR__ . '/fixtures/sites/framablog.test'))
+        );
+        $client = new Client();
+        $client->getEmitter()->attach(new Mock([$response]));
+
+        $graby = new Graby();
+        $res = $graby->fetchContent('https://framablog.org/2017/12/02/avancer-ensemble-vers-la-contribution/');
+        // The initial treatment was encapsulating the content into the empty node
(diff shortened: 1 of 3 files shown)


j0k3r:
Could you rebase against the master?
And also, add a test to cover that case (maybe using the framablog link in the `GrabyFunctionalTest.php`)

j0k3r:
You can let the call happen :)
But move it to `GrabyFunctionalTest`, you'll see a lot of real test case targeting existing website in that file.

tests/GrabyFunctionalTest.php
```diff
@@ -583,4 +583,25 @@ public function testJsonLd()
...
+    {
+        $response = new Response(
+            200,
+            [
+                'content-type' => 'text/html',
+            ],
+            Stream::factory(file_get_contents(__DIR__ . '/fixtures/sites/framablog.test'))
+        );
+        $client = new Client();
+        $client->getEmitter()->attach(new Mock([$response]));
```
j0k3r: You can remove the mock, it is not used and it breaks the build: https://travis-ci.org/j0k3r/graby/jobs/388131182#L490

Simounet:
I rolled back but the mock isn't working anyway. :/

tests/GrabyTest.php
```diff
@@ -1436,4 +1436,23 @@ public function testEncodingUtf8ForTextPlainPage()
...
     }
+
+    public function testEmptyNodesRemoved()
+    {
+        $response = new Response(
+            200,
+            [
+                'content-type' => 'text/html',
+            ],
+            Stream::factory(file_get_contents(__DIR__ . '/fixtures/sites/framablog.test'))
```
j0k3r: I don't think the `file_get_contents` is required here.

j0k3r:
Not yet :slightly_smiling_face: 
I'm wondering if your changes doesn't break the iframe on lemonde test

Simounet:
Same player, play again! :D

PR #158, opened 2018-07-11
Meta name author auto detection

Simounet:
Hi there,
Because it is [standard stuff](https://www.w3.org/TR/html5/document-metadata.html#standard-metadata-names), I think it's a good idea to add that to the auto detection feature through the `<meta name="author" content="Mario">` tag.
Have a nice day.


Simounet:
Done.

j0k3r:
Could you check for the failing test?

Simounet:
I checked but there is more than one problem about things I didn't changed. :/

j0k3r:
Looks like it's the new test which fail, not a new one.
https://travis-ci.org/j0k3r/graby/jobs/436707931#L558

Also, I've rebased your branch and fixed an unrelated test on the go.

PR #166, opened 2018-10-04
Use of name from JsonLd as title

Simounet:
We can use the `name` from the JsonLd (https://schema.org/VideoObject). Not sure about `headline` pertinence as a title.


j0k3r:
Question is: what's the priority between headline & name? Which one should override the other one?

Simounet:
As far I understood, name if more important than headline.

j0k3r:
:+1:

PR #168, opened 2018-10-18
Add Additional Guzzle configs

eudaco:
With Additional Guzzle configs you can set proxy settings or other Curl parameters.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "j0k3r/graby",
+    "name": "eudaco/graby",
     "description": "Graby helps you extract article content from web pages",
     "license": "MIT",
     "authors": [{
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "j0k3r/graby",
+    "name": "eudaco/graby",
```
j0k3r: I can't merge the PR with this kind of change

src/Extractor/HttpClient.php
```diff
@@ -65,6 +65,9 @@ public function __construct(Client $client, $config = [], LoggerInterface $logge
             'timeout' => 10,
             // number of redirection allowed until we assume request won't be complete
             'max_redirect' => 10,
+			
+			// Additional Guzzle configs.
+			'config' => []
```
j0k3r: This additional config is too vague to be merged.
Maybe we can define a more appropriate config, like, I don't know, `proxy_settings`, etc.

j0k3r:
This PR is fairly old and there hasn't been much activity on it. Closing, but please re-open when you'll work on it.

PR #169, opened 2018-10-24
Switch to official tcpdf

j0k3r:
https://nvd.nist.gov/vuln/detail/CVE-2018-17057

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,15 +25,17 @@
         "symfony/options-resolver": "~2.6|~3.0",
         "monolog/monolog": "^1.13.1",
         "smalot/pdfparser": "~0.11",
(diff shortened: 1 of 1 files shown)


jtojnar:
Why not update our fork?

j0k3r:
I just updated _our_ fork and I see that we don't have much difference: https://github.com/tecnickcom/TCPDF/compare/master...wallabag:master

The only difference is related `$maxchwidth` but the way we fixed it isn't really a fix because the variable won't be declared after and might (also) generate a warning later.

I don't know if the bug still exist in the official repo.
We should check with https://github.com/wallabag/wallabag/issues/2776

j0k3r:
I've updated our fork, we'll see later to make the change.

PR #178, opened 2018-11-21
Update Graby.php (regex modification)

techexo:
I propose this regex modification to actually clean empty nodes before using php-readability and/or Tidy.
This regex specifically exclude the first group to be an `iframe` (which was the intention according to the comment), but allows the deletion of empty attribute-containing nodes.

With this modification, will be detected and deleted the following examples:
```html
<h2></h2>
<p></p>
<h2 class="align"></h2>
<head></head>
```
but not
```html
<iframe></iframe>
<iframe src="xxx.html"></iframe>
```

You can test it here: https://regex101.com/r/uWjsOc/1

It fixes https://github.com/wallabag/wallabag/issues/3730 and make https://github.com/j0k3r/php-readability/issues/39 closable (as I had no proof it was a _php-readability_ issue).

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -254,7 +254,7 @@ class Graby
         $html = $this->convert2Utf8($response['body'], $response['all_headers']);
 
         // Remove empty nodes (except iframe)
-        $re = '/<(?!iframe)([^>\s]+)(?:[^>]*)>(?:\s*(?:<br \/>|&nbsp;|&thinsp;|&ensp;|&emsp;|&#8201;|&#8194;|&#8195;)\s*)*<\/\1>/m';
+        $re = '/<(?!iframe)([^>\s]+)[^>]*>(?:\s*(?:<br \/>|&nbsp;|&thinsp;|&ensp;|&emsp;|&#8201;|&#8194;|&#8195;)\s*)*<\/\1>/m';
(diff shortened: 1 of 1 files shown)


techexo:
**Actually, someone might want to check that it is not a very bad idea!**
In the present form, it will also delete `<script src="..."></script>` instructions in header (but I think it is stripped anyway by php-readability). As far as I know, other instructions in header are most of the time "mono-markup" (e.g. `<meta .../>`, `<link ... />`)

techexo:
It seems that my code breaks quite a lot of tests, I will have a look and propose a modification if I can find what went wrong.

j0k3r:
First rebase against the master. I might have fixed some of failing tests (which might not be related to your changes)

techexo:
I rebased, but it was a _me_ issue, not a _test_ issue ;)
I was allowing for `>` in extra characters, so consecutive markups were squashed together... It should be OK now with the new commit!

techexo:
@j0k3r I analyzed a bit more the original regex, and I am pretty confident that it doesn't discriminate _at all_ the `iframe` with the group exclusion `[^iframe|>]`, but with the fact that _iframes_ markups always have a `src=""` following.
I have commited the simpler version mentionned above.

PR #183, opened 2018-12-04
Regex update to strip empty nodes containing spaces

techexo:
The precedent regex couldn't detect if there was only "normal" spaces
inside empty nodes. I think some groups were wrongfully intertwined
sometime in the past.

You can check here to see examples/tests:
https://regex101.com/r/YdPShz/1/


j0k3r:
Great!

PR #184, opened 2018-12-10
Debug mode for graby

techexo:
Hello,

Here is my attempt at a better debug mode for Graby. The idea is to fully use the various levels of information possible that can be passed to the logger (`Logger::DEBUG` and `Logger::INFO`) to separate log messages. Graby's config is changed to `'debug' => false` to `'log_level' => false`, and expect `debug` or `info` as argument.

I chose the `Logger::INFO` level to store what was previously in the debug mode, as it is indeed information about the steps taken, without too much details. The `Logger::DEBUG` level stores in a separate log file `html.log` the data at various steps during the fetching/cleaning process. _Note that I put something to empty this file each time graby's called._

By default, nothing is stored ; if `'log_level' => 'info'`, only INFO messages are stored in `graby.log` ; if `'log_level' => 'debug'`, INFO messages are in `graby.log` and `html.log` contains debug messages of the last call to graby.

Most of the work is between lines 74-87 of Graby.php, but I also modified the logging calls to use the aliases offered by Monolog (i.e. `$this->logger->debug()` instead of `$this->logger->log('debug', ...)`). This is negotiable, I just found it nicer.

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -147,7 +147,7 @@ class ContentExtractor
         $configFingerprint = $this->configBuilder->buildForHost($fingerprintHost);
 
         if (!empty($this->config['fingerprints']) && false !== $configFingerprint) {
-            $this->logger->log('debug', 'Appending site config settings from {host} (fingerprint match)', ['host' => $fingerprintHost]);
+            $this->logger->info('Appending site config settings from {host} (fingerprint match)', ['host' => $fingerprintHost]);
             $this->configBuilder->mergeConfig($config, $configFingerprint);
 
             if ($addToCache && false === $this->configBuilder->getCachedVersion($fingerprintHost)) {
@@ -184,31 +184,36 @@ class ContentExtractor
             $this->config['src_lazy_load_attributes'][] = $this->siteConfig->src_lazy_load_attr;
         }
 
+        $this->logger->debug('Actual site config', ['siteConfig' => $this->siteConfig]);
+
         // do string replacements
         if (!empty($this->siteConfig->find_string)) {
             if (\count($this->siteConfig->find_string) === \count($this->siteConfig->replace_string)) {
                 $html = str_replace($this->siteConfig->find_string, $this->siteConfig->replace_string, $html, $count);
-                $this->logger->log('debug', 'Strings replaced: {count} (find_string and/or replace_string)', ['count' => $count]);
+                $this->logger->info('Strings replaced: {count} (find_string and/or replace_string)', ['count' => $count]);
             } else {
-                $this->logger->log('debug', 'Skipped string replacement - incorrect number of find-replace strings in site config');
+                $this->logger->info('Skipped string replacement - incorrect number of find-replace strings in site config');
             }
             unset($count);
         }
+
+        $this->logger->debug('HTML after site config strings replacements', ['html' => $html]);
 
         // load and parse html
         $parser = $this->siteConfig->parser();
 
         if (!\in_array($parser, $this->config['allowed_parsers'], true)) {
-            $this->logger->log('debug', 'HTML parser {parser} not listed, using {default_parser} instead', ['parser' => $parser, 'default_parser' => $this->config['default_parser']]);
+            $this->logger->info('HTML parser {parser} not listed, using {default_parser} instead', ['parser' => $parser, 'default_parser' => $this->config['default_parser']]);
             $parser = $this->config['default_parser'];
         }
 
-        $this->logger->log('debug', 'Attempting to parse HTML with {parser}', ['parser' => $parser]);
+        $this->logger->info('Attempting to parse HTML with {parser}', ['parser' => $parser]);
 
         $this->readability = $this->getReadability($html, $url, $parser, $this->siteConfig->tidy() && $smartTidy);
         $tidied = $this->readability->tidied;
 
-        $this->logger->log('debug', 'Body size after Readability: {length}', ['length' => \strlen($this->readability->dom->saveXML())]);
+        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen($this->readability->dom->saveXML())]);
+        $this->logger->debug('Body after Readability', ['dom_saveXML' => $this->readability->dom->saveXML()]);
 
         // we use xpath to find elements in the given HTML document
         $this->xpath = new \DOMXPath($this->readability->dom);
@@ -246,7 +251,7 @@ class ContentExtractor
 
         // try to get title
         foreach ($this->siteConfig->title as $pattern) {
-            $this->logger->log('debug', 'Trying {pattern} for title', ['pattern' => $pattern]);
+            $this->logger->info('Trying {pattern} for title', ['pattern' => $pattern]);
 
             if ($this->extractEntityFromPattern('title', $pattern)) {
                 break;
@@ -256,7 +261,7 @@ class ContentExtractor
         // try to get author (if it hasn't already been set)
         if (empty($this->authors)) {
             foreach ($this->siteConfig->author as $pattern) {
-                $this->logger->log('debug', 'Trying {pattern} for author', ['pattern' => $pattern]);
+                $this->logger->info('Trying {pattern} for author', ['pattern' => $pattern]);
 
                 if ($this->extractMultipleEntityFromPattern('authors', $pattern)) {
                     break;
@@ -266,7 +271,7 @@ class ContentExtractor
 
         // try to get date
         foreach ($this->siteConfig->date as $pattern) {
-            $this->logger->log('debug', 'Trying {pattern} for date', ['pattern' => $pattern]);
+            $this->logger->info('Trying {pattern} for date', ['pattern' => $pattern]);
 
             if ($this->extractEntityFromPattern('date', $pattern)) {
                 break;
@@ -276,13 +281,13 @@ class ContentExtractor
         // try to get language
         $langXpath = ['//html[@lang]/@lang', '//meta[@name="DC.language"]/@content'];
         foreach ($langXpath as $pattern) {
-            $this->logger->log('debug', 'Trying {pattern} for language', ['pattern' => $pattern]);
+            $this->logger->info('Trying {pattern} for language', ['pattern' => $pattern]);
             $elems = $this->xpath->evaluate($pattern, $this->readability->dom);
 
             if ($elems instanceof \DOMNodeList && $elems->length > 0) {
                 foreach ($elems as $elem) {
                     $this->language = trim($elem->textContent);
-                    $this->logger->log('debug', 'Language matched: {language}', ['language' => $this->language]);
+                    $this->logger->info('Language matched: {language}', ['language' => $this->language]);
                 }
 
                 if (null !== $this->language) {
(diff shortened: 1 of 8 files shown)


techexo:
Wooops, it seems that I troubled a few things by deleting the variable `$debug`! I will try to see how I can get around that, by redefining it when log_level = debug (or info? I don't know what would be better).

techexo:
Actually, it is not so much as a logging issue, but the tests are not valid anymore as they call Graby with `'debug' => true` and we would need to have `'log_level' => 'info'` instead.
Furthermore, maybe [this statement](https://github.com/techexo/graby/blob/53b6f127b97c0d03a12ee0359b2b33279357a39e/src/Graby.php#L413) would have to be in the INFO stream and not in the DEBUG stream if the tests use it.

I don't want to try and fix the tests for now, I prefer to see first how this PR is welcomed 😄 !

techexo:
@j0k3r I finally modified the tests to take into account the new loglevels and rebased everything on master as of today.

src/Graby.php
```diff
@@ -43,7 +43,7 @@ public function __construct($config = [], Client $client = null, ConfigBuilder $
     {
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
-            'debug' => false,
```
j0k3r: Removing a configuration mean it'll break backward compatibility.
Why don't you want to keep `debug`?
- debug = true, we want verbose log
- debug = false, we don't want verbose log

Or if you want to keep different log, introduce a new log level configuration, with default value to "info" and when it's "debug", give more log.

src/Extractor/ContentExtractor.php
```diff
@@ -184,31 +184,36 @@ public function process($html, $url, SiteConfig $siteConfig = null, $smartTidy =
...
 
-        $this->logger->log('debug', 'Attempting to parse HTML with {parser}', ['parser' => $parser]);
+        $this->logger->info('Attempting to parse HTML with {parser}', ['parser' => $parser]);
 
         $this->readability = $this->getReadability($html, $url, $parser, $this->siteConfig->tidy() && $smartTidy);
         $tidied = $this->readability->tidied;
 
-        $this->logger->log('debug', 'Body size after Readability: {length}', ['length' => \strlen($this->readability->dom->saveXML())]);
+        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen($this->readability->dom->saveXML())]);
+        $this->logger->debug('Body after Readability: ', ['dom_saveXML' => $this->readability->dom->saveXML()]);
```
```suggestion
        $this->logger->debug('Body after Readability', ['dom_saveXML' => $this->readability->dom->saveXML()]);
```

src/Extractor/ContentExtractor.php
```diff
@@ -355,9 +360,11 @@ public function process($html, $url, SiteConfig $siteConfig = null, $smartTidy =
 
         $this->removeElements($elems, 'Stripping {length} empty a elements');
 
+        $this->logger->debug('DOM after site config stripping: ', ['dom_saveXML' => $this->readability->dom->saveXML()]);
```
```suggestion
        $this->logger->debug('DOM after site config stripping', ['dom_saveXML' => $this->readability->dom->saveXML()]);
```

src/Graby.php
```diff
@@ -229,6 +237,8 @@ public function cleanupHtml($contentBlock, $url)
             $html = preg_replace('!</?a[^>]*>!', '', $html);
         }
 
+        $this->logger->debug('Body after cleanupHtml, before cleanupXss:', ['html' => $html]);
```
j0k3r:
```suggestion
        $this->logger->debug('Body after cleanupHtml, before cleanupXss', ['html' => $html]);
```

src/Graby.php
```diff
@@ -263,6 +273,8 @@ private function doFetchContent($url)
 
         $html = $this->convert2Utf8($response['body'], $response['all_headers']);
 
+        $this->logger->debug('Fetched HTML:', ['html' => $html]);
```
j0k3r:
```suggestion
        $this->logger->debug('Fetched HTML', ['html' => $html]);
```

src/Graby.php
```diff
@@ -272,6 +284,8 @@ private function doFetchContent($url)
         $re = '/<(?!iframe)([^>\s]+)[^>]*>(?:<br \/>|&nbsp;|&thinsp;|&ensp;|&emsp;|&#8201;|&#8194;|&#8195;|\s)*<\/\1>/m';
         $html = preg_replace($re, '', $html);
 
+        $this->logger->debug('HTML after regex empty nodes stripping:', ['html' => $html]);
```
j0k3r:
```suggestion
        $this->logger->debug('HTML after regex empty nodes stripping', ['html' => $html]);
```

src/Extractor/ContentExtractor.php
```diff
@@ -184,31 +184,36 @@ public function process($html, $url, SiteConfig $siteConfig = null, $smartTidy =
             $this->config['src_lazy_load_attributes'][] = $this->siteConfig->src_lazy_load_attr;
         }
 
+        $this->logger->debug('Actual site config: ', ['siteConfig' => $this->siteConfig]);
```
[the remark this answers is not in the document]
techexo: Removed with https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

src/Extractor/ContentExtractor.php
```diff
@@ -184,31 +184,36 @@ public function process($html, $url, SiteConfig $siteConfig = null, $smartTidy =
...
-                $this->logger->log('debug', 'Strings replaced: {count} (find_string and/or replace_string)', ['count' => $count]);
+                $this->logger->info('Strings replaced: {count} (find_string and/or replace_string)', ['count' => $count]);
             } else {
-                $this->logger->log('debug', 'Skipped string replacement - incorrect number of find-replace strings in site config');
+                $this->logger->info('Skipped string replacement - incorrect number of find-replace strings in site config');
             }
             unset($count);
         }
 
+        $this->logger->debug('HTML after site config strings replacements: ', ['html' => $html]);
```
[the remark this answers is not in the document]
techexo: Resolved in https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

src/Extractor/ContentExtractor.php
techexo: Resolved in https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

src/Graby.php
techexo: Resolved in https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

src/Graby.php
techexo: Resolved in https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

src/Graby.php
techexo: Resolved in https://github.com/j0k3r/graby/pull/184/commits/0a7d2ca09e7a0f81d9d66ac24b4acad48990d09e

fetcher.php
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/php
+<?php
+
+require 'vendor/autoload.php';
```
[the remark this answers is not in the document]
techexo: Yup, I made a mistake while pushing :) Solved with https://github.com/j0k3r/graby/pull/184/commits/88e15690aac5e6873db565d5b4b85df1803f8e64

src/Graby.php
techexo: I am not sure I understand what you mean. I get it about backward compatibility, but what would you want instead?

The issue with keeping just `debug = true/false` is that I cannot introduce the `Logger::INFO` level easily in the if/else statement. In the current situation, when `debug = true`, the equivalent to my `log_level = info` is logged in graby.log ; else nothing is logged with `debug = false`.

I think that most people won't use the "full debug", so I wouldn't want to activate it by default and give the choice between the current debug mode and the more detailed one. So I'd need a way to still use two levels of Logger, with one variable. Do you have an idea in mind?

techexo: Cf commit https://github.com/j0k3r/graby/pull/184/commits/eca074a8029860119ccbdc3642029866a84ff5b5

techexo:
@j0k3r The last commits should resolve all your requested changes. I had to change the definition of the TestLogger (to get rid of Logger::DEBUG level) for content extractor's tests. Also I realized that I didn't modify logging in HttpClient.php, this is now done.

I realize that the place where I added Logger::DEBUG messages were very subjective, and mostly to help me debug site configuration files on wallabag. These commits are just to put in place an architecture for better logging.

src/Graby.php
```diff
@@ -70,12 +72,25 @@ public function __construct($config = [], Client $client = null, ConfigBuilder $
...
+            if ($this->logLevel == 'debug') {
+                // Emptying of the HTML logfile to avoid gigantic logs
+                // This statement has to be before Logger::INFO to catch all DEBUG messages
+                fclose(fopen(__DIR__ . '/../log/html.log','w'));
+                $this->logger->pushHandler(new StreamHandler(__DIR__ . '/../log/html.log', Logger::DEBUG));
+            }
+
+            $this->logger->pushHandler(new StreamHandler(__DIR__ . '/../log/graby.log', Logger::INFO, false));
+        } else {
+            $this->logger = new NullLogger();
```
j0k3r: You can avoid the `else` by letting the definition before the `if` (like it was done before)

tests/GrabyTest.php
```diff
@@ -17,7 +17,7 @@ public function testConstructDefault()
     {
         $graby = new Graby(['debug' => true]);
 
-        $this->assertTrue($graby->getConfig('debug'));
```
j0k3r: You should keep this one too

src/Graby.php
```diff
@@ -70,12 +72,25 @@ public function __construct($config = [], Client $client = null, ConfigBuilder $
...
         $this->config = $resolver->resolve($config);
 
-        $this->debug = (bool) $this->config['debug'];
-        $this->logger = new NullLogger();
+        $this->debug = $this->config['debug'];
+        $this->logLevel = $this->config['log_level'];
 
+        // Debug mode can be activated with 'debug' => true
+        // More details on the retrieved code and its consequent modifications
+        // can be obtained using 'log_level' = 'debug'
```
j0k3r: That line can be at the end of the previous one, it won't be a _too long_ comment on one line

tests/GrabyTest.php
techexo: https://github.com/j0k3r/graby/pull/184/commits/df26846ea7690e95bf00a445d0be095e799884eb

src/Graby.php
techexo: https://github.com/j0k3r/graby/pull/184/commits/df26846ea7690e95bf00a445d0be095e799884eb

j0k3r:
> Edit: there is an issue with Travis last PHP7.2 tests, but I am not sure I'm at fault for it, could you have a look?

Run locally:

```
php vendor/bin/php-cs-fixer fix
```

j0k3r:
Strange it did well on my local machine.
I've squashed your commits and it'll be ready to merge :rocket:

PR #187, opened 2019-01-04
Fix wallabag#3825

techexo:
Delete empty lines in retrieved HTML code to avoid runaway evaluation of empty node stripping regex on badly coded websites.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -265,8 +265,8 @@ class Graby
 
         // Remove empty lines to avoid runaway evaluation of following regex
         // on badly coded websites
-        $re = '/(^[\r\n]*|[\r\n]+)[\s\t]*[\r\n]+/';
-        $html = preg_replace($re, '\n', $html);
+        $re = '/^[ \t]*[\r\n]+/m';
+        $html = preg_replace($re, '', $html);
 
         // Remove empty nodes (except iframe)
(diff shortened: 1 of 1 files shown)


techexo:
Note: unit tests are failing (as could be expected) because of code not having empty lines any more.

techexo:
Sorry, did not see your review. I am absolutely not comfortable with unit tests (I know what it is, but never wrote one, especially in PHP...).
What would you like it to test? Runaway evaluation or just line stripping?

j0k3r:
> Note: unit tests are failing (as could be expected) because of code not having empty lines any more.

Maybe because it might not be the best way to fix it? I didn't took time to investigate just asking.

techexo:
I am trying a new regex right now, which might be less of a problem. (And sorry for spamming with commits, but I don't know how to run my test locally, so I push and wait for Github's Travis to do the job :smirk:...).

In any way, I think that from the beginning going at the HTML code with regex is a call for future disaster (as eloquently put in _php-readability_ : _HACK: dirty cleanup to replace some stuff; shouldn't use regexps with HTML but well..._).

Edit: It seems that the new regex allows for fixing the issue without breaking the tests.

j0k3r:
It reminds me that post on SO https://stackoverflow.com/q/1732348/569101 :face_with_head_bandage:

techexo:
It would be interesting to see if it's possible to improve graby & php-readability to get rid of regexes (except for replacements in site configuration files). I'll try to have a go at it someday.

PR #192, opened 2019-01-21
Avoid bad date to be send

j0k3r:
If the date can't be parsed, it means it's a bad one and shouldn't be return.

For example https://www.helpnetsecurity.com/2015/07/22/its-official-the-average-ddos-attack-size-is-increasing/ has:

```html
<meta property="article:published_time" content="-0001-11-30T00:00:00+00:00" />
```


src/Extractor/ContentExtractor.php
```diff
@@ -531,6 +531,15 @@ function ($element, $currentEntity) {
         // If no year has been found during date_parse, we nuke the whole value
         // because the date is invalid
         if (false === $parseDate['year']) {
+            $this->logger->info('Date is bad (wrong year): {date}', ['date' => $this->date]);
+
+            $this->date = null;
+        }
+
+        // in case the date can't be converted we assume it's a wrong date
+        if (0 > strtotime($this->date)) {
```
tcitworld: `strtotime` returns `false` since PHP 5.1 :)

j0k3r: But `strtotime('-0001-11-30T00:00:00+00:00')` return a negative value: https://3v4l.org/P40gB
But yeah, I should check for `false` too

j0k3r:
@tcitworld the xpath can return whatever we want. But I think graby shouldn't return a date that can't be parsed as a date

PR #199, opened 2019-03-22
lesswrong.com workaround

Najrim:
See comment on commit. Based on issue [#3767](https://github.com/wallabag/wallabag/issues/3767) in Wallabag repo.

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -38,6 +38,7 @@ class HttpClient
                 'tnr.com' => ['tnr.com/article/' => 'tnr.com/print/article/'],
                 '.m.wikipedia.org' => ['.m.wikipedia.org' => '.wikipedia.org'],
                 'm.vanityfair.com' => ['m.vanityfair.com' => 'www.vanityfair.com'],
+                'www.lesswrong.com' => ['www.lesswrong.com' => 'www.greaterwrong.com'],
             ],
             // Prevent certain file/mime types
             // HTTP responses which match these content types will


techexo:
I am interested to see if this PR will be accepted, because some other websites would require such a workaround. I think @j0k3r didn't want to do such modifications in the core code of graby.

Maybe there is a way to include them not in the core source code, but in graby's configuration on wallabag ([maybe here?](https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Resources/config/services.yml). I think only @j0k3r will be able to tell us.

j0k3r:
Yeah that's a good question.
I would like to work on a way to use a configuration to add or remove this kind of `rewrite_url` pattern.
Or maybe this should goes into site config instead of hard coded in the coded.
A configuration of graby for that might be a good idea but what if an other website than wallabag wants the same behavior for `lesswrong`? It'll have to add it inside the configuration of graby too.

What do you think @fivefilters?

fivefilters:
Yes, I agree @j0k3r. I think it'd be better if it was handled in the site config file. I've been thinking about this too as there are sites where adding a parameter saves an uncessary redirect or avoids a GDPR/cookie wall from being displayed. So perhaps in addition to rewrite_url, we could also have a way to add/remove query string parameters.

Najrim:
Alright, there seems to be consensus that this should be in a configuration file instead so I will close the pull request for now. My reason for proposing the change is just personal convenience. I'm not self-hosting at the moment so I can't fix this locally.

PR #214, opened 2019-10-04
Add brand new logo to the project

caneco:
Following your request in the tweet:
https://twitter.com/j0k/status/1180008533619888128

Here it is your brand new logo. **Hope you like it!**

image


caneco:
TBH… I like the idea that the "G" can "grab" something… but I'm not a truly fan of the thing, so here's another option.

TBH… I like the idea that the "G" can "grab" something… but not a trully

Maybe you can play of the holding G for other stuff?

jtojnar:
Wow, I love the hand idea, and the colours too. To make it grab something like an article would be nice touch but I cannot imagine doing it in a way that would not completely destroy the logo.

caneco:
@jtojnar sorry the misspell… I had the same idea for the G to be holding an article, but it maybe confuse the thing. Going to try it anyway 😃

j0k3r:
Nice job!
I agree with @jtojnar about the G grabbing an article. Let’s see what you can try ☺️

caneco:
Hello again… before I update the PR, what do you think about this approach?


image

j0k3r:
I prefer that one.
Can we merge the G with the rest of the word, so it’s not Ggraby? ☺️

jtojnar:
Could you also check how does it look like on a dark backgroung? I use https://github.com/StylishThemes/GitHub-Dark and the white background of the current image is little jarring:

image

j0k3r:
@caneco any news? :)

caneco:
Hello, sorry I totally miss left this aside :shame:

So, regarding your comments here are mine:

> Can we merge the G with the rest of the word, so it’s not Ggraby?
It's better not… either way could mislead people to read "RABY" 😕
Sorry, but need to blame the "holding article" 😅… that's on me
image

> Could you also check how does it look like on a dark backgroung?
On black, actually it would be better to have your suggestion above but without the "holding article"
image
but all in white…

image


**Comments?**

jtojnar:
I am rather fond of purple. Could we keep that even on dark background?

Other than that SVG variant should be committed to the repo and a license should be decided. Do we want to use Graby’s MIT, or use something like CC or Art Libre like Wallabag does for [its logo](https://github.com/wallabag/logo).

caneco:
@jtojnar meaning the one without holding the article?

caneco:
> @jtojnar meaning the one without holding the article?
do you?

@jtojnar && @j0k3r regarding the SVG version I'm postponing that subject to November, according to the [CoC](https://www.notion.so/Do-you-need-a-logo-14b46a005e184aaca3481f722d765a90) released

j0k3r:
No problem for the SVG version later.
I like the second one, with the `G` size more aligned with `raby`. I think the final version can be this one with the text/card inside the G.

As a bonus, can we see how the colorful logo render with a black background?

caneco:
Like this?

image
(on white)

image
(on black)

j0k3r:
Yup like this 👍 
The SVG (or the PNG) will have to have a transparent background and it'll be ok on both white & black background.

You can update the PR

caneco:
Just updated the PR! Go for it!!! 🎉

jtojnar:
Looks superbly.

However, would you mind moving the credits to the bottom of the README like [selfoss](https://github.com/SSilence/selfoss#credits) or [Wallabag](https://github.com/wallabag/wallabag#license) do? It is important to provide attribution but, in my opinion, the README is meant to quickly help the user find out, in this order:

1. What the repo is about?
2. How to make it run?
3. How to use it?
4. How to contribute?

The order is designed to copy the stages user will enter when interacting with the software. That means, that it is not appropriate to put the credits at the top, since it would delay user from achieving their goal.

And since we do not have the option of overlaying the credits over the content like TV series have, we can only create a separate section at the end:

5. Who is behind it?

j0k3r:
@jtojnar don't worry, I'll update the README later :)

PR #216, opened 2019-10-07
Convert date to a known format

j0k3r:
Instead of just validating the date is ok when parsing it, just convert it to the W3C format

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,9 @@ before_install:
 install:
     - composer up --no-interaction --prefer-dist -o
 
+before_script:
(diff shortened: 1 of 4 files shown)


jtojnar:
Why not just make `validateDate` return `?DateTime`?

j0k3r:
@jtojnar because it'll break compatibility with current date return. Today we return a plain text which is a valid date according to `parse_date` function. Now it'll return a formatted date instead but it'll still be a string.

tcitworld:
`DateTimeInterface` ? Available since PHP 7.2. Looks odd.

j0k3r:
Nope, it's only the constant that are move to `DateTimeInterface` in PHP 7.2, before that they were defined in `DateTime`:

> 7.2.0 | The class constants of DateTime are now defined on DateTimeInterface.

PR #220, opened 2020-01-06
Lock to PHPUnit 7

j0k3r:
PHPUnit 8 require PHP >= 7.2
Graby require PHP >= 7.1

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,7 +5,7 @@ includes:
 parameters:
     # https://github.com/phpstan/phpstan/issues/694#issuecomment-350724288
     autoload_files:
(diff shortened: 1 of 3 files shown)


Kdecherf:
As PHP 7.1 reached EOL a few weeks ago, could we define a date for sunsetting its support in Graby?

tcitworld:
After wallabag 2.4 release? :p

jtojnar:
The fact that php.net no longer supports a certain release is not relevant for many people, as they often use their distribution’s fork of PHP, which has security issues fixed by e.g. RedHat. As before, I would suggest creating a new major branch when dropping old versions of PHP, so that selfoss can switch to that (we target PHP used by [supported Debian versions](https://wiki.debian.org/PHP#Available_versions).

PR #225, opened 2020-03-29
Remove the wallabag's fork of tcpdf

Kdecherf:
This PR removes the explicit dependency to wallabag/tcpdf as graby does not use it itself and it may prevent wallabag from moving back to the former package (`tecnickcom/TCPDF`).


j0k3r:
In fact, it was used to override the version used in `smalot/pdfparser`.

PR #244, opened 2020-12-11
Add `VideoGame` JsonLD type to exceptions

zyuhel:
Some resources like Gamespot.com have VideoGame JsonLD objects that available on the page and because they are last, they rewrite correct datatime with incorrect one.

Example: 
https://www.gamespot.com/articles/nvidia-rtx-3090-3080-3070-gpu-all-specs-prices-and-release-dates-detailed/1100-6481673/

p.s Maybe there is a good idea to make this list customizable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ configure it and inject it to `Graby\Graby`.
 This is the full documented configuration and also the default one.
 
 ```php
-$graby = new Graby(array(
+$graby = new Graby([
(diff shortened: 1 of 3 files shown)


j0k3r:
Thanks @zyuhel! Sorry for the year of delay 😅
I've updated your PR and added an option to `ContentExtractor` called `json_ld_ignore_types`.

PR #246, opened 2020-12-27
Rebuild siteconfig if host for singlepage is not the same than the so…

Kdecherf:
Consider a `single_page_link` directive on twitter.com that follows any
link found in a tweet. Here is an example with the following tweet:

  https://twitter.com/Cloudflare/status/1341353044504694787

This eventually leads to the following page:

  https://blog.cloudflare.com/beat-an-acoustics-inspired-ddos-attack/

Until now, Graby was incorrectly sending to cloudflare.com
http headers defined for twitter.com.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -745,7 +745,19 @@ class Graby
         // check it's not what we have already!
         if (false !== $singlePageUrl && $singlePageUrl !== $url) {
             // it's not, so let's try to fetch it...
-            $response = $this->httpClient->fetch($singlePageUrl, false, $siteConfig->http_header);
+            $headers = $siteConfig->http_header;
+
+            $sourceUrl = parse_url($url);
+            $targetUrl = parse_url($singlePageUrl);
+            if (\is_array($sourceUrl)
+                && \is_array($targetUrl)
+                && \array_key_exists('host', $sourceUrl)
+                && \array_key_exists('host', $targetUrl)
+                && $sourceUrl['host'] !== $targetUrl['host']) {
+                $targetSiteConfig = $this->configBuilder->buildForHost($targetUrl['host']);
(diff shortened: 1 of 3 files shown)


j0k3r (commented):
Interesting!
Could you fix PHPStan & add some tests?

src/Graby.php
```diff
@@ -745,7 +745,19 @@ private function getSinglePage($html, $url)
...
+            $sourceUrl = parse_url($url);
+            $targetUrl = parse_url($singlePageUrl);
+            if (\is_array($sourceUrl)
+                && \is_array($targetUrl)
+                && \array_key_exists('host', $sourceUrl)
+                && \array_key_exists('host', $targetUrl)
+                && $sourceUrl['host'] !== $targetUrl['host']) {
+                $targetSiteConfig = $this->configBuilder->buildForHost($targetUrl['host']);
+                $headers = $targetSiteConfig->http_header;
+            }
```
j0k3r, line 759: When using multiline in the `if` condition, I usually move the opening brackets on a new line so it's easier to see when the condition list end and the code start.

```suggestion
            if (\is_array($sourceUrl)
                && \is_array($targetUrl)
                && \array_key_exists('host', $sourceUrl)
                && \array_key_exists('host', $targetUrl)
                && $sourceUrl['host'] !== $targetUrl['host'])
            {
                $targetSiteConfig = $this->configBuilder->buildForHost($targetUrl['host']);
                $headers = $targetSiteConfig->http_header;
            }
```

Kdecherf: Noted, I'll push a fixup

Kdecherf: php-cs-fixer is reporting an issue with this change

j0k3r (approved):

PR #248, opened 2021-01-10
ContentExtractor: fix handling of data-srcset

Kdecherf:
The extractor was incorrectly setting the value of data-srcset into the
src attribute, however srcset does not use the same format leading to
loading error with incorrect paths. Now data-srcset sets the value of
srcset which will let the browser to use paths of either src or srcset
attributes.

This PR may require new tests to ensure non-regression.

Here is an example with `https://www.numerama.com/sciences/661291-covid-19-le-modele-emmental-montre-aussi-pourquoi-la-desinformation-est-si-grave.html`: 

```
<img class="size-large wp-image-661339" alt="" src="https://www.numerama.com/wp-content/uploads/2020/11/sgr-1935-2154-sursauts-radio-rapides-1024x576.jpg%201024w,%20https://www.numerama.com/content/uploads/2020/10/gestes_barrieres_emmental-680x383.jpg%20680w,%20https://www.numerama.com/content/uploads/2020/10/gestes_barrieres_emmental-1536x864.jpg%201536w,%20https://www.numerama.com/content/uploads/2020/10/gestes_barrieres_emmental.jpg%201920w" referrerpolicy="no-referrer" width="1024" height="576">
```

Fixes https://github.com/wallabag/wallabag/issues/4914

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -607,16 +607,23 @@ class ContentExtractor
                     continue;
                 }
 
-                $src = null;
+                $attributes = [];
                 foreach ($this->config['src_lazy_load_attributes'] as $attribute) {
                     if ($e->hasAttribute($attribute)) {
-                        $src = $e->getAttribute($attribute);
+                        $key = 'src';
+                        if ('data-srcset' === $attribute) {
+                            $key = 'srcset';
+                        }
+                        $attributes[$key] = $e->getAttribute($attribute);
                         $e->removeAttribute($attribute);
                     }
                 }
 
-                if (null !== $src) {
-                    $e->setAttribute('src', $src);
+                foreach (['src', 'srcset'] as $attr) {
+                    if (\array_key_exists($attr, $attributes)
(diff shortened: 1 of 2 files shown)


src/Extractor/ContentExtractor.php
```diff
@@ -607,16 +607,24 @@ function ($element, $currentEntity) {
...
                         $e->removeAttribute($attribute);
                     }
                 }
 
-                if (null !== $src) {
-                    $e->setAttribute('src', $src);
+                foreach (['src', 'srcset'] as $attr) {
+                    if (\is_array($attributes)
+                            && \array_key_exists($attr, $attributes)
+                            && null !== $attributes[$attr]) {
```
j0k3r, line 625: I think `$attributes` is always an array at this point. Don't you think?

```suggestion
                    if (\array_key_exists($attr, $attributes) && null !== $attributes[$attr]) {
```

Kdecherf: Indeed, I'll push a fixup

j0k3r (approved):

PR #252, opened 2021-01-22
HttpClient: refactor the way we clean utm_ query params

Kdecherf:
The old `preg_replace` call incorrectly removed the `?` of the query string
section which could lead to things like this:

  `http://example.com/foo?utm_source=a&var=value` =>
`http://example.com/foo&var=value`

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -109,7 +109,7 @@ class HttpClient
         );
     }
 
-    public function setLogger(LoggerInterface $logger)
+    public function setLogger(LoggerInterface $logger): void
     {
         $this->logger = $logger;
     }
@@ -282,12 +282,8 @@ class HttpClient
 
     /**
      * Cleanup URL and retrieve the final url to be called.
-     *
-     * @param string $url
-     *
-     * @return string
      */
-    private function cleanupUrl($url)
+    private function cleanupUrl(string $url): string
     {
         // rewrite part of urls to something more readable
         foreach ($this->config['rewrite_url'] as $find => $action) {
@@ -323,10 +319,8 @@ class HttpClient
      * by checking the extension.
      *
      * @param string $url Absolute url
-     *
-     * @return bool
      */
-    private function possibleUnsupportedType($url)
+    private function possibleUnsupportedType(string $url): bool
     {
         $ext = strtolower(trim(pathinfo($url, \PATHINFO_EXTENSION)));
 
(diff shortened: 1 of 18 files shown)


src/Extractor/HttpClient.php
```diff
@@ -263,7 +263,12 @@ public function fetch($url, $skipTypeVerification = false, $httpHeader = [])
         }
 
         // remove utm parameters & fragment
-        $effectiveUrl = preg_replace('/((\?)?(&(amp;)?)?utm_(.*?)\=[^&]+)|(#(.*?)\=[^&]+)/', '', rawurldecode($effectiveUrl));
```
jtojnar, line 266: Interestingly, the regex also handles `&amp;`-separated queries. Not sure why it did that. One possible sources (`getMetaRefreshURL`) decodes them:

https://github.com/j0k3r/graby/blob/e1154098851dd347002314445c18847655f5a289/src/Extractor/HttpClient.php#L529

jtojnar: Maybe `getUglyURL`?

Kdecherf: I don't think so, `getUglyURL` and `getMetaRefreshURL` calls, should they return anything, would lead to the exit of the method before reaching my change:

https://github.com/j0k3r/graby/blob/e1154098851dd347002314445c18847655f5a289/src/Extractor/HttpClient.php#L257-L263

However @j0k3r is there any reason to have handled `&amp;` here?

j0k3r: I think it was just in case we have that kind of given url

Kdecherf: So should we continue to handle them?

j0k3r: I think so, is it hard to achieve?

j0k3r: FullTextRSS seems to have `preg_replace('/(\?|\&)utm_[a-z]+=[^\&]+/', '', $url);`

That line in graby was added during the first commit (👴), I've no idea why I've added the `&amp;`.

Maybe you just avoid a regression, you can do:

```php
$uri = new Uri(str_replace('&amp;', '&', $effectiveUrl));
```

Kdecherf: I've included your suggested change @j0k3r

j0k3r (commented):
Looks great, you can squash :+1:

j0k3r:
Sh*t I fucked up the PR when changing the base branch and then merged it 🤦‍♂️
Could you reopen the same PR targeting the `2.x` branch without my latest changes from the upcoming 3.0?
So your PR will be part of the next 2.x release and also 3.0.
Thanks and sorry for the extra work..

PR #255, opened 2021-04-09
Case where an attribute is empty overriding a filled in value

Simounet:
Hi there,
Fix a case where an empty attribute were overriding a filled in one.
Have a nice day and stay safe.


Simounet:
I removed the assert verifying the language attribute is null because it's what I want to avoid. Using an empty attribute is not helpful, is it?

Simounet:
Sorry. It's an issue with the master branch not from my PR. I'm trying to fix it anyway.

j0k3r:
Thanks!

PR #256, opened 2021-04-13
Fix wrong URL exception handling

Simounet:
Hi there,
I encountered a really ugly URL (from a wrong editor copy/paste I guess) that makes Graby fail. I fixed it and made the test case.


src/Graby.php
```diff
@@ -815,7 +815,12 @@ private function makeAbsoluteAttr($base, \DOMNode $e, $attr)
         $url = trim(str_replace('%20', ' ', $e->getAttribute($attr)));
         $url = str_replace(' ', '%20', $url);
 
-        $absolute = $this->makeAbsoluteStr($base, $url);
+        try {
+            $absolute = $this->makeAbsoluteStr($base, $url);
+        } catch (\Exception $exception) {
+            $this->logger->info('Wrong content url', ['url' => $url]);
+            $absolute = $url;
```
j0k3r, line 822: If we failed to make it absolute, shouldn't we skip it instead?

```suggestion
            $absolute = false;
```

Simounet: I think we should keep this URL that the user can see the issue. Plus, we should keep the content, even if it's bad content. :/

PR #257, opened 2021-04-13
Fix test case with real content (bjori site updated)

Simounet:
I updated the test case but I was wondering why you were fetching real site instead of using existing fixture?


j0k3r:
Thanks for fixing it.
I fetch a real site because I want to test a real scenario instead of mocking stuff.
And sometimes site are updated and test fail.

PR #274, opened 2021-12-04
Add ability to process prefetched content

Kdecherf:
fetchContent() now accepts an optional parameter, prefetchedContent, which
can contain the content of a page that was fetched outside of Graby.

If we take the example of Wallabag it gives the ability to send the
content of a page (_through a browser extension for example_) without
making network calls to fetch the actual page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,22 @@ array(
 
 The `date` result is the same as displayed in the content. If `date` is not `null` in the result, we recommend you to parse it using [`date_parse`](http://php.net/date_parse) (this is what we are using to validate that the date is correct).
 
+### Retrieve content from a prefetched page
+
+If you want to extract content from a page you fetched outside of Graby, you can call `setContentAsPrefetched()` before calling `fetchContent()`, e.g.:
+
+``` php
+use Graby\Graby;
+
+$article = 'http://www.bbc.com/news/entertainment-arts-32547474';
+
+$input = '<html>[...]</html>';
+
+$graby = new Graby();
+$graby->setContentAsPrefetched($input);
+$result = $graby->fetchContent($article);
+```
+
 ### Cleanup content
 
 Since the 1.9.0 version, you can also send html content to be cleanup in the same way graby clean content retrieved from an url. The url is still needed to convert links to absolute, etc.
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -34,6 +34,8 @@ class Graby
     private Punycode $punycode;
     private bool $imgNoReferrer = false;
 
+    private ?string $prefetchedContent = null;
+
     public function __construct(array $config = [], Client $client = null, ConfigBuilder $configBuilder = null)
     {
         $this->config = new GrabyConfig($config);
@@ -101,6 +103,11 @@ class Graby
     public function reloadConfigFiles(): void
     {
         $this->configBuilder->loadConfigFiles();
+    }
+
+    public function setContentAsPrefetched(string $content): void
+    {
+        $this->prefetchedContent = $content;
     }
 
     /**
@@ -208,6 +215,18 @@ class Graby
         return trim($this->cleanupXss((string) $html));
     }
 
+    private function getResponseForPrefetchedContent(string $url): array
+    {
+        return [
+            'body' => $this->prefetchedContent,
+            'effective_url' => $url,
+            'headers' => [
+                'content-type' => 'text/html',
+            ],
+            'status' => 200,
+        ];
+    }
+
     /**
      * Do fetch content from an url.
      *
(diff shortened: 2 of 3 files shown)


jtojnar:
Could you please add some context – what is the purpose of this? If it is for tests, would not it be cleaner to mock `httpClient`?

Kdecherf:
@jtojnar I worked a few weeks ago on a WebScrapBook server implementation in wallabag, it lets me save actual page content from my browser using the extension [WebScrapBook](https://github.com/danny0838/webscrapbook) which is useful for pages with javascript rendering or bot protection.

This implies to send content to Graby without it to make a HTTP call, thus adding support for "prefetched content".

Kdecherf:
This PR is ready for review I think, poke @jtojnar @j0k3r

src/Graby.php
```diff
@@ -167,11 +167,11 @@ public function getConfig(string $key)
      *
      * @return array With keys html, title, url & summary
      */
-    public function fetchContent(string $url): array
+    public function fetchContent(string $url, string $prefetchedContent = null): array
```
jtojnar, line 170: Passing a content to a method called `fetchContent` still feels a bit iffy to me, I would prefer a separate `cleanContent` method. But then, the code has been a pile of megafunctions playing fast and loose with SRP since it was called Full-Text RSS, and I do not see a way to separate it out cleanly without a huge refactoring.

src/Graby.php
```diff
@@ -322,7 +338,7 @@ private function doFetchContent(string $url): array
 
         // check site config for single page URL - fetch it if found
         $isSinglePage = false;
-        if ($this->config['singlepage'] && ($singlePageResponse = $this->getSinglePage($html, $effectiveUrl))) {
+        if ($this->config['singlepage'] && null === $prefetchedContent && ($singlePageResponse = $this->getSinglePage($html, $effectiveUrl))) {
```
jtojnar, line 341: Would not it be reasonable to pass content from web browser but still want Graby to collect multi-page data? Wallabag can always set `singlepage = false` and `multipage = false` to prevent this.

src/Graby.php
Kdecherf: Separating fetch and extraction logics from `doFetchContent()` may be difficult because of the multipage fetcher.

Another approach instead of passing content in the `fetchContent()` could be to have a separate setter like `setContentAsPrefetched($content)` that we would call before calling `fetchContent()`.

src/Graby.php
Kdecherf: The issue is that I don't see how the http `fetch()` call occurring on lines 407 and 757 could work when you provide a prefetched content.

We could consider that the caller (wallabag in my case) should set singlepage=false and multipage=false when providing prefetched content, however there is no way to set these variables as "oneshot operations" afaik.

src/Graby.php
j0k3r: Great suggestion @Kdecherf  with `setContentAsPrefetched`. I prefer it. Can you make the change?

Kdecherf: @j0k3r done, I think we're good to make another review

j0k3r (commented):
Looks good to me, could you add an entry in the readme about that new behavior?

Kdecherf:
> Looks good to me, could you add an entry in the readme about that new behavior?

@j0k3r https://github.com/j0k3r/graby/pull/274/commits/7aa0592609ddfaabaa69f5efb5788b09f154fbbc do you think it's enough?

README.md
```diff
@@ -129,6 +129,23 @@ array(
...
+
+$input = '<html>[...]</html>';
+
+$graby = new Graby();
+$graby->setContentAsPrefetched($input);
+$result = $graby->fetchContent($article);
+
+```
+
 The `date` result is the same as displayed in the content. If `date` is not `null` in the result, we recommend you to parse it using [`date_parse`](http://php.net/date_parse) (this is what we are using to validate that the date is correct).
```
j0k3r, line 132: That text should be above your content

README.md
```diff
@@ -129,6 +129,23 @@ array(
...
+use Graby\Graby;
+
+$article = 'http://www.bbc.com/news/entertainment-arts-32547474';
+
+$input = '<html>[...]</html>';
+
+$graby = new Graby();
+$graby->setContentAsPrefetched($input);
+$result = $graby->fetchContent($article);
+
```
j0k3r, line 146:
```suggestion
```

README.md
Kdecherf: Woops, fixed

j0k3r (commented):
Go squash! 👍

j0k3r (approved):

PR #276, opened 2021-12-11
HttpClient: refactor the way we clean utm_ query params

Kdecherf:
PR for 2.x, see https://github.com/j0k3r/graby/pull/252#issuecomment-991308070


j0k3r (approved):
Thanks!

Issue #278, opened 2022-01-07
Getting 502 Bad gateway due to PHP Errors/Warnings


girishpanchal30:
Hello,

I'm getting 502 bad gateway errors due to graby library. I have listed PHP warnings/errors below. It is generating multiple-time PHP errors and due to this, we have a memory leaked issue in our server.

`[Wed Jan 05 13:33:08.565297 2022] [php7:warn] [pid 12003] [client 70.39.235.45:43632] PHP Warning:  DOMDocument::saveXML(): unknown encoding  in /vendor/j0k3r/graby/src/Extractor/ContentExtractor.php on line 380`

**Feed URL: http://feeds.feedburner.com/motherjones/feed**

`PHP Fatal error:  Uncaught TypeError: Typed property Graby\\Extractor\\ContentExtractor::$title must be string or null, array used in /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php:1414\nStack trace:\n#0 /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php(1225): Graby\\Extractor\\ContentExtractor->extractJsonLdInformation()\n#1 /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php(244): Graby\\Extractor\\ContentExtractor->extractDefinedInformation()\n#2 /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Graby.php(344): Graby\\Extractor\\ContentExtractor->process()\n#3 /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Graby.php(174): Graby\\Graby->doFetchContent()\n#4 /var/www/html/wp-content/plugins/feedzy-full-post/classes/FeedzyFullPost/Parser/Library/Graby.php(32): Graby\\Graby->fetchContent()\n#5 /var/www/html/wp-content/plugins/feedzy-ful in /var/www/html/wp-content/plugins/feedzy-full-post/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php on line 1414`

**Feed URL: https://news.google.com/news/rss/search/section/q/pakistan covid/pakistan covid?**

[...]

PR #279, opened 2022-01-07
Fix PHP Error/Warnning

girishpanchal30:
- PHP Warning:  array_combine(): Both parameters should have an equal number

- PHP Warning:  DOMDocument::saveXML(): unknown encoding

- PHP Fatal error:  Uncaught TypeError: Typed property Graby\\Extractor\\ContentExtractor::$title must be string or null

issue - #278

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -201,8 +201,8 @@ class ContentExtractor
         $this->readability = $this->getReadability($html, $url, $parser, $this->siteConfig->tidy() && $smartTidy);
         $tidied = $this->readability->tidied;
 
-        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen((string) $this->readability->dom->saveXML())]);
-        $this->logger->debug('Body after Readability', ['dom_saveXML' => $this->readability->dom->saveXML()]);
+        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen((string) $this->readability->dom->saveXML($this->readability->dom->documentElement))]);
+        $this->logger->debug('Body after Readability', ['dom_saveXML' => $this->readability->dom->saveXML($this->readability->dom->documentElement)]);
 
         // we use xpath to find elements in the given HTML document
         $this->xpath = new \DOMXPath($this->readability->dom);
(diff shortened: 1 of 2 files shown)


src/SiteConfig/ConfigBuilder.php
```diff
@@ -327,6 +327,11 @@ public function mergeConfig(SiteConfig $currentConfig, SiteConfig $newConfig)
         //      find_string: <other-img
         //      replace_string: <img
         // To fix that issue, we combine find & replace as key & value in one array, we merge them and then rebuild find & replace string in the current config
+
+        if (count($currentConfig->find_string) !== count($currentConfig->replace_string)) {
+            return $currentConfig;
+        }
+
```
j0k3r, line 334: Could you then move the `http_header` definition before that new condition?

girishpanchal30: moved `http_header` definition

src/SiteConfig/ConfigBuilder.php
```diff
@@ -340,7 +346,6 @@ public function mergeConfig(SiteConfig $currentConfig, SiteConfig $newConfig)
             $currentConfig->replace_string[] = $replaceString;
         }
 
-        // merge http_header array from currentConfig into newConfig
         // because final values override former values in case of named keys
         $currentConfig->http_header = array_merge($newConfig->http_header, $currentConfig->http_header);
```
j0k3r, line 350: I mean, you should move the whole 3 lines

girishpanchal30: Done

j0k3r:
Looks great, you can squash your commits :+1:

j0k3r:
Thanks @girishpanchal30!

src/SiteConfig/ConfigBuilder.php
```diff
@@ -327,6 +327,15 @@ public function mergeConfig(SiteConfig $currentConfig, SiteConfig $newConfig)
...
         //      replace_string: <img
         // To fix that issue, we combine find & replace as key & value in one array, we merge them and then rebuild find & replace string in the current config
+
+        // merge http_header array from currentConfig into newConfig
+        // because final values override former values in case of named keys
+        $currentConfig->http_header = array_merge($newConfig->http_header, $currentConfig->http_header);
+
+        if (\count($currentConfig->find_string) !== \count($currentConfig->replace_string)) {
+            return $currentConfig;
+        }
```
jtojnar, line 337: This looks weird to me. Sohuld not we rather return `$newConfig`? And should not we check the variables from `$newConfig`? I would assume the `$currentConfig` has been already validated upon construction.

src/Extractor/ContentExtractor.php
```diff
@@ -201,8 +201,8 @@ public function process(string $html, string $url, SiteConfig $siteConfig = null
         $this->readability = $this->getReadability($html, $url, $parser, $this->siteConfig->tidy() && $smartTidy);
         $tidied = $this->readability->tidied;
 
-        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen((string) $this->readability->dom->saveXML())]);
-        $this->logger->debug('Body after Readability', ['dom_saveXML' => $this->readability->dom->saveXML()]);
+        $this->logger->info('Body size after Readability: {length}', ['length' => \strlen((string) $this->readability->dom->saveXML($this->readability->dom->documentElement))]);
```
jtojnar, line 204: If https://www.php.net/manual/en/domdocument.savexml.php#44422 is correct, this would just save a page from unknown encoding as UTF-8. No idea how would it try to do that, possibly leading to some corruption.

j0k3r: You're right, but that's not really a problem as it's inside the log, maybe the log will only be unreadable?

src/SiteConfig/ConfigBuilder.php
j0k3r: You're right, I'll move that check in `parseLines` instead

PR #280, opened 2022-01-07
Ensure empty value from `site_config` are kept

j0k3r:
For example, when user wants to replace a string with nothing, `replace_string:` will be used and shouldn't be skipped.
Also, adding more tests regarding [previous PR](https://github.com/j0k3r/graby/pull/279).

### DIFF
--- a/tests/GrabyFunctionalTest.php
+++ b/tests/GrabyFunctionalTest.php
@@ -340,4 +340,36 @@ class GrabyFunctionalTest extends TestCase
         // if the cookie wasn't taking into account, it'll be "NPR Choice page"
         $this->assertSame('Michael Flynn\'s Contradictory Line On Russia', $res['title']);
     }
+
+    public function testSaveXmlUnknownEncoding(): void
+    {
+        $graby = new Graby([
+            'debug' => true,
+            'extractor' => [
+                'config_builder' => [
+                    'site_config' => [__DIR__ . '/fixtures/site_config'],
+                ],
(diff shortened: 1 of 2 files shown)


tests/GrabyFunctionalTest.php
```diff
@@ -340,4 +340,36 @@ public function testCookie(): void
...
+    {
+        $graby = new Graby([
+            'debug' => true,
+            'extractor' => [
+                'config_builder' => [
+                    'site_config' => [__DIR__ . '/fixtures/site_config'],
+                ],
+            ],
+        ]);
+        $res = $graby->fetchContent('http://motherjones.com/politics/2012/02/mac-mcclelland-free-online-shipping-warehouses-labor');
```
jtojnar, line 354: I wonder if we could store the response for mocking purposes in case the site ever gets down.

j0k3r: I started that for few cases but it's really boring to perform 😫
https://github.com/j0k3r/graby/tree/master/tests/fixtures/content

jtojnar: I guess I could try making an automatic rule for https://github.com/rectorphp/rector.

j0k3r: You mean a rule that run the test suite and save content fetched in file?

jtojnar: Opened #282.

PR #281, opened 2022-01-07
Ensure `find_string` & `replace_string` size are equal

j0k3r:
That validation is performed in `parseLines` so the size of them can't mismatch when merging configuration for example.
A warning log is raised in case of mismatch.

Following https://github.com/j0k3r/graby/pull/279#discussion_r780230930
Poke @jtojnar

### DIFF
--- a/src/SiteConfig/ConfigBuilder.php
+++ b/src/SiteConfig/ConfigBuilder.php
@@ -332,11 +332,6 @@ class ConfigBuilder
         //      replace_string: <img
         // To fix that issue, we combine find & replace as key & value in one array, we merge them and then rebuild find & replace string in the current config
 
-        // in case of bad configuration
-        if (\count($currentConfig->find_string) !== \count($currentConfig->replace_string)) {
-            return $currentConfig;
-        }
-
         $findReplaceCurrentConfig = array_combine($currentConfig->find_string, $currentConfig->replace_string);
         $findReplaceNewConfig = array_combine($newConfig->find_string, $newConfig->replace_string);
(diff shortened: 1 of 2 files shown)


src/SiteConfig/ConfigBuilder.php
```diff
@@ -412,6 +407,14 @@ public function parseLines(array $lines)
             }
         }
 
+        // in case of bad configuration
+        if (\count($config->find_string) !== \count($config->replace_string)) {
+            $this->logger->warning('find_string & replace_string size mismatch, check the site config to fix it', ['find_string' => $config->find_string, 'replace_string' => $config->replace_string]);
+
+            $config->find_string = [];
+            $config->replace_string = [];
```
jtojnar, line 415: Maybe we could zip them together (i.e. trim them to the length of the shorter)?

j0k3r: Won't we have undesired side effects?

In that case, it'll work
```
find_string: badstuff
replace_string: goodone
replace_string: src="
```

In that case, it'll not work because you'll replace `hellow` by `goodone` if you _zip_ them.
```
find_string: hellow
find_string: badstuff
replace_string: goodone
```

I think it's preferable to empty them and log a warning to the user instead.

jtojnar (approved):

src/SiteConfig/ConfigBuilder.php
jtojnar: Either way sounds reasonable to me.

PR #282, opened 2022-01-07
Add test fixtures using rector

jtojnar:
Use `composer rector` (or `composer rector -- --dry-run`) to run the automated refactoring.


j0k3r:
Maybe you can separate the code improvement (`fn`, `??`, etc.) in a separate PR to reduce the noise?

jtojnar:
I can separate them out but then CI will not pass.

jtojnar:
Or I can move the ci commit to the new PR too.

jtojnar:
Removed the ci part for now.

Still need to replace `file_get_contents` with actual `fetchContent` and capture the resources.

jtojnar:
I got nerd sniped quite hard by this but I had this on to-do list for few other projects. Now running `composer rector` should automatically add fixtures for `fetchContent` calls on `Graby` constructed without a second argument.

j0k3r (commented):
I don't get everything about Rector (never used it).
What I understand if I add a new functional test is I'll need to run it once and then the fixture html will be generated and I'll need to update the functional test to use the fixture.
Am I right ?
If so, how do you catch headers you added in the test file?

jtojnar:
Rector is a tool for automated code transformation. The transformation rule runs Graby and extract headers using `RecordingHttpClient` will also update the functional tests – the last commit is fully produced by the transformation rule without any manual changes.

jtojnar:
If you check out the PR, you can try it yourself by removing the last commit and running `composer rector`. (comment out [this line](https://github.com/jtojnar/graby/blob/a2d0269fd3060f84978b60fadc0b85ec5dfa9776/maintenance/rector.config.php#L12) if you do not want the code modernization).

tests/GrabyFunctionalTest.php
```diff
@@ -312,14 +328,17 @@ public function testMultipage(): void
 
     public function testCookie(): void
     {
+        $httpMockClient = new HttpMockClient();
+        $httpMockClient->addResponse(new Response(301, ['Server' => ['AkamaiGHost'], 'Content-Length' => ['0'], 'Location' => ['https://www.npr.org/sections/parallels/2017/05/19/529148729/michael-flynns-contradictory-line-on-russia'], 'Date' => ['Sat, 08 Jan 2022 05:02:29 GMT'], 'Connection' => ['keep-alive']], (string) file_get_contents(__DIR__ . '/fixtures/content/http___www.npr.org_sections_parallels_2017_05_19_529148729_michael-flynns-contradictory-line-on-russia.html')));
```
jtojnar, line 332: Perhaps we should update the link to remove the redirect.

tests/GrabyFunctionalTest.php
```diff
@@ -312,14 +328,17 @@ public function testMultipage(): void
...
+        $httpMockClient->addResponse(new Response(200, ['Content-Type' => ['text/html;charset=UTF-8'], 'X-Cache-NPR' => ['MISS'], 'Access-Control-Allow-Origin' => ['http://www.google.co.uk'], 'Access-Control-Allow-Credentials' => ['true'], 'X-NPR-Trace-ID' => ['renIaLEEV1Q'], 'X-Content-Type-Options' => ['nosniff'], 'X-XSS-Protection' => ['1; mode=block'], 'X-Served-By' => ['pod-www-render-nginx-776fbf4c8c-7sjbx'], 'Referrer-Policy' => ['no-referrer-when-downgrade'], 'Strict-Transport-Security' => ['max-age=15724800; includeSubDomains'], 'Cache-Control' => ['no-cache'], 'Expires' => ['Sat, 08 Jan 2022 05:02:30 GMT'], 'Date' => ['Sat, 08 Jan 2022 05:02:30 GMT'], 'Transfer-Encoding' => ['chunked'], 'Connection' => ['keep-alive', 'Transfer-Encoding']], (string) file_get_contents(__DIR__ . '/fixtures/content/http___www.npr.org_sections_parallels_2017_05_19_529148729_michael-flynns-contradictory-line-on-russia')));
```
jtojnar, line 333: I guess this will not actually do anything if we want to send cookie.

j0k3r: You're right, could you add an exclusion rule for Rector to not auto mock that request?

jtojnar: There are three options for this:

- Have blacklist of function names
- Use some in code marker (e.g. a comment like `Rector: do not add mocking`
- Pass `null` as second argument (with a comment for humans that we do not want mocking) – this one works now.

What would you prefer.

j0k3r: The code marker is an annotation to the function to blacklist?

j0k3r: I'm ok with that

jtojnar: It is now skipped.

j0k3r:
Can you rebase against the master?
And it'll be good to merge

j0k3r:
Good job @jtojnar 👊

PR #283, opened 2022-01-11
Code update thanks to Rector

j0k3r:
Thanks to @jtojnar in https://github.com/j0k3r/graby/pull/282

Also:
- cast some variable to improve PHPStan
- catch `JsonException` when parsing JSON-LD

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -25,22 +25,10 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
  */
 class HttpClient
 {
-    /**
-     * @var array
-     */
-    private $config;
-    /**
-     * @var HttpMethodsClient
-     */
-    private $client;
-    /**
-     * @var LoggerInterface
-     */
-    private $logger;
-    /**
-     * @var History
-     */
-    private $responseHistory;
+    private array $config;
+    private HttpMethodsClient $client;
+    private LoggerInterface $logger;
+    private History $responseHistory;
 
     /**
      * @param Client $client Http client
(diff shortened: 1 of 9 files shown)


src/Extractor/HttpClient.php
```diff
@@ -245,7 +233,7 @@ public function fetch($url, $skipTypeVerification = false, $httpHeader = [])
         // (regex found here: https://stackoverflow.com/a/137831/569101)
         preg_match_all('/<!--\[if\s(?:[^<]+|<(?!!\[endif\]-->))*<!\[endif\]-->/mi', $body, $matchesConditional);
 
-        if (isset($matchesConditional[0]) && \count($matchesConditional[0]) > 1) {
+        if (isset($matchesConditional[0]) && (is_countable($matchesConditional[0]) ? \count($matchesConditional[0]) : 0) > 1) {
```
jtojnar, line 236: Hmm, I would expect `$matchesConditional` to always be array of arrays so it should be countable: https://www.php.net/manual/en/function.preg-match-all.php Maybe it is Rector/PHPStan bug.

src/Extractor/HttpClient.php
```diff
@@ -543,7 +527,7 @@ private function getUglyURL(string $url, string $html)
     {
         $found = false;
         foreach ($this->config['ajax_triggers'] as $string) {
-            if (stripos($html, $string)) {
+            if (stripos($html, (string) $string)) {
```
jtojnar, line 530: We should either use value object with typed properties or otherwise validate the config when passed to Graby. Then it would not add pointless casts like this.

j0k3r: I would like to implement https://symfony.com/doc/current/components/options_resolver.html#type-validation

src/Extractor/HttpClient.php
j0k3r: I guess so

PR #284, opened 2022-01-24
Convert configuration to Value Object

j0k3r:
Mostly to enforce type of each parameter of a config.
Also, enforce type for the input configuration using `setAllowedTypes`.

See https://github.com/j0k3r/graby/pull/283#pullrequestreview-849144036

Is it something like that you were talking about @jtojnar?

### DIFF
--- a/src/Extractor/ContentExtractorConfig.php
+++ b/src/Extractor/ContentExtractorConfig.php
@@ -48,6 +48,7 @@ class ContentExtractorConfig
         ]);
 
         $resolver->setAllowedValues('default_parser', $this->allowed_parsers);
+
         $resolver->setAllowedTypes('default_parser', 'string');
         $resolver->setAllowedTypes('fingerprints', 'array');
         $resolver->setAllowedTypes('config_builder', 'array');
--- a/src/Extractor/ContentExtractorConfig.php
+++ b/src/Extractor/ContentExtractorConfig.php
@@ -13,9 +13,9 @@ class ContentExtractorConfig
 {
     use ArrayStringOptionsTrait;
 
+    private const ALLOWED_PARSERS = ['libxml', 'html5lib'];
+
     private string $default_parser;
-    /** @var array<string> */
-    private array $allowed_parsers = ['libxml', 'html5lib'];
     /** @var array<string, string> */
     private array $fingerprints;
     private array $config_builder;
@@ -51,7 +51,7 @@ class ContentExtractorConfig
             ],
         ]);
 
-        $resolver->setAllowedValues('default_parser', $this->allowed_parsers);
+        $resolver->setAllowedValues('default_parser', self::ALLOWED_PARSERS);
 
         $resolver->setAllowedTypes('default_parser', 'string');
         $resolver->setAllowedTypes('fingerprints', 'array');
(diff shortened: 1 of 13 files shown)


jtojnar:
Yeah, that looks about right. Once we bump to PHP 8, we will even be able to switch to [named parameters](https://php.watch/versions/8.0/named-parameters) and ditch [options resolver](https://symfony.com/doc/current/components/options_resolver.html) altogether. And then with [`readonly` properties](https://php.watch/versions/8.1/readonly), we won’t even need the getters.

j0k3r:
I think I'm done @jtojnar, can you take a look? 🙏

src/Extractor/ContentExtractorConfig.php
```diff
@@ -0,0 +1,115 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Configuration for ContentExtractor as a Value Object.
+ */
+class ContentExtractorConfig
+{
+    private string $default_parser;
+    /** @var array<string> */
+    private array $allowed_parsers = ['libxml', 'html5lib'];
```
jtojnar, line 18: Maybe make this `const` if it is not allowed to change?

tests/GrabyTest.php
```diff
@@ -897,26 +862,6 @@ public function testContentLinksRemove(): void
         $this->assertFalse($res['native_ad']);
     }
 
-    public function testMimeActionNotDefined(): void
```
jtojnar, line 900: Why was this removed? Should be probably mentioned in the commit message.

j0k3r: I removed it because it's now checked by the OptionResolver, so the behavior changed (ie: it's not more a silent error): https://github.com/j0k3r/graby/blob/0c7cd9375225a9476365384990e3deabf2f70a1f/src/GrabyConfig.php#L74

jtojnar: Oh, right. I guess that can have `array<string, array{name: string, action: 'link'|'exclude'}>` annotation: https://phpstan.org/writing-php-code/phpdoc-types

src/GrabyConfig.php
```diff
@@ -0,0 +1,165 @@
...
+    private string $error_message_title;
+    /** @var array<string> */
+    private array $allowed_urls;
+    /** @var array<string> */
+    private array $blocked_urls;
+    private bool $xss_filter;
+    private array $content_type_exc;
+    private string $content_links;
+    private array $http_client;
+    private array $extractor;
```
jtojnar, line 28: Then we might have objects here.

src/SiteConfig/ConfigBuilder.php
```diff
@@ -29,22 +28,9 @@ class ConfigBuilder
         'div',
     ];
 
-    /**
-     * @param array $config
-     */
-    public function __construct($config = [], LoggerInterface $logger = null)
+    public function __construct(array $config = [], LoggerInterface $logger = null)
```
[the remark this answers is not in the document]
j0k3r: I agree it could allow to better check for sub-configuration but I found it more verbose. You'll have to _use_ object config for each level. I think it'll add complexity where array is really simple to use.

PR #290, opened 2022-03-01
Fix cookies on multiple page

### LINKED ISSUE #48: multipage articles do not reuse cookie?
E.g. https://www.golem.de/news/app-entwicklung-cross-platform-oder-nativ-programmieren-2202-162600.html

When i add this article, i do not get the content of pages 2 and following. Instead of the content from page 2, i see a notice about the missing cookie consent. But this is defined in https://github.com/j0k3r/graby-site-config/blob/master/golem.de.txt#L5 already. Is the cookie sent as header when fetching consecutive pages of an article?
Fetching page 1 works fine.

j0k3r:
It was the case for golem.de, the cookie wasn't properly send to the next page (might be a bug in the cookie jar not properly retrieving previous defined cookies).

Fix https://github.com/j0k3r/graby-site-config/issues/48

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,7 @@
 parameters:
     level: 7
     paths:
-        - maintenance/Rector
+        #- maintenance/Rector
         - src
         - tests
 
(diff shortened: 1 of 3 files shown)


Kdecherf:
Didn't check in depth but we may need to add a test to ensure that we don't leak cookies if the next page is not on the same domain. What do you think?

j0k3r:
That's a good question, this shouldn't happen so much.
Should we add the cookie in the cookie jar instead so it'll be checked later in the foreach?

j0k3r:
I've checked and cookies we are re-injecting are only those which are defined in site config, so there shouldn't really contains sensitive data?
Otherwise I don't know how to fix the _leak_ because at the point where we should check it, we don't know the host associated to cookies which are in the headers.

PR #293, opened 2022-06-11
rector: Update for latest version

jtojnar:
Currently broken for some reason:

If I revert the test fixtures, `composer rector` is no longer able to add them with Rector ≥ 0.13.1. It works with 0.13.0 and older. I do not see anything suspicious in the changes though https://github.com/rectorphp/rector-src/compare/0.13.0...0.13.1.

Opened https://github.com/rectorphp/rector/issues/7231.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         }
     },
     "scripts": {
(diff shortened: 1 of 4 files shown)


j0k3r:
@jtojnar Thanks!

PR #309, opened 2023-03-06
Fix CS & upgrade GitHub Actions

j0k3r:
Long time without an update and it requires an update. Also:
- use `http_build_query` to rebuild the query instead of `Uri::withQueryValues` because the array from the filter method might contains array and it's not supported (reported by PHPStan and validated with a new test)
- fix deprecating of `\SimplePie_Misc`

Replace https://github.com/j0k3r/graby/pull/308

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -220,10 +220,7 @@ class HttpClient
         }
 
         // remove utm parameters & fragment
-        $uri = new Uri(str_replace('&amp;', '&', $effectiveUrl));
-        parse_str($uri->getQuery(), $query);
-        $queryParameters = array_filter($query, fn ($k) => !(0 === stripos((string) $k, 'utm_')), \ARRAY_FILTER_USE_KEY);
-        $effectiveUrl = (string) new Uri($uri->withFragment('')->withQuery(http_build_query($queryParameters)));
+        $effectiveUrl = (string) $this->removeTrackersFromUrl(new Uri(str_replace('&amp;', '&', $effectiveUrl)));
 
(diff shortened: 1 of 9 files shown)


src/Extractor/HttpClient.php
```diff
@@ -222,8 +222,8 @@ public function fetch($url, $skipTypeVerification = false, $httpHeader = [])
         // remove utm parameters & fragment
         $uri = new Uri(str_replace('&amp;', '&', $effectiveUrl));
         parse_str($uri->getQuery(), $query);
-        $queryParameters = array_filter($query, fn ($k) => !(0 === stripos($k, 'utm_')), \ARRAY_FILTER_USE_KEY);
-        $effectiveUrl = (string) Uri::withQueryValues(new Uri($uri->withFragment('')->withQuery('')), $queryParameters);
+        $queryParameters = array_filter($query, fn ($k) => !(0 === stripos((string) $k, 'utm_')), \ARRAY_FILTER_USE_KEY);
+        $effectiveUrl = (string) new Uri($uri->withFragment('')->withQuery(http_build_query($queryParameters)));
```
jtojnar, line 226: Note that `http_build_query` is not an inverse of `parse_str`:

```php
php > parse_str('foo[]=x', $query); var_dump(http_build_query($query));
string(12) "foo%5B0%5D=x"
```

jtojnar: In selfoss, I have decided to just use `explode` for this purpose:

https://github.com/fossar/selfoss/blob/0d7bde56e502f7d79bfb38dcdd657c7da89cf1f1/src/spouts/rss/fulltextrss.php#L120

jtojnar (approved):

PR #310, opened 2023-03-06
Graby: Clarify returned array value types for fetchContent

jtojnar:
This will allow stricter PHPStan checking. Though ideally we would replace arrays with objects.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -109,7 +109,7 @@ class Graby
     /**
      * Fetch content from the given url and return a readable content.
      *
-     * @return array With keys html, title, url & summary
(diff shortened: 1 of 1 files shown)


src/Graby.php
```diff
@@ -226,7 +226,7 @@ private function getResponseForPrefetchedContent(string $url): array
     /**
      * Do fetch content from an url.
      *
-     * @return array With key status, html, title, language, date, authors, url, image, headers & native_ad
+     * @return array{status: int, html: string, title: string, language: ?string, date: ?string, authors: string[]|null, url: string, image: ?string, headers: array<string, string>, native_ad: bool}
```
jtojnar, line 229: Looks like author array can contain `false`:

https://github.com/j0k3r/graby/blob/7de23d400791c5c0e343d26475e483519da03338/src/Extractor/ContentExtractor.php#L1338

jtojnar:
Superseded by #316.

PR #311, opened 2023-03-07
Remove slow Scrutinizer 👋

j0k3r:


jtojnar:
There is also a mention in gitignore.

Issue #302, opened 2022-11-06
function convert2Utf8() return wrong


ducwp:
I tried with this url: https://www.24h.com.vn/tin-tuc-trong-ngay/dieu-tra-vu-chi-hang-ty-dong-tiep-khach-cua-van-phong-hdnd-tinh-gia-lai-c46a1412042.html.

It get content with character encoding error.

Then I open file **j0k3r/graby/Graby.php**, add `$html = $response['body']`  after line 326 (dont use **convert2Utf8** ), it's work fine.

![dontuseutf](https://user-images.githubusercontent.com/18102229/200157070-77d564e5-d756-4a01-8849-688c81ed76fb.png)

ducwp:
Or if I add `$response['body'] = str_replace('<meta charset = "UTF-8">', '<meta charset="UTF-8">', $response['body']);` before line 326, it will work fine.

![Screenshot_21](https://user-images.githubusercontent.com/18102229/200157259-64850c0b-2194-43a2-adc9-08861a30e331.png)

PR #312, opened 2023-03-07
Graby::convert2Utf8: Allow spaces around HTML attributes

jtojnar:
HTML attributes can contain a spaces around the equal sign: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

Fixes: https://github.com/j0k3r/graby/issues/302


j0k3r (approved):
Good job!

PR #315, opened 2023-03-29
ContentExtractor: clean up validateDate

jtojnar:
- Drop incorrect and redundant phpdoc param hint.
- Terminate early on null so that we can drop later casts and null checks.
- Add a comment about suspicious cast introduced in https://github.com/j0k3r/graby/commit/6480564f8ff01122f288af6dc90523297a4b5f50
- Catch a possible Exception caused by DateTime constructor.

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -685,31 +685,39 @@ class ContentExtractor
     /**
      * Validate and convert a date to the W3C format.
      *
-     * @param string $date
-     *
      * @return string|null Formatted date using the W3C format (Y-m-d\TH:i:sP) OR null if the date is badly formatted
      */
     public function validateDate(?string $date): ?string
     {
-        $date = (string) $date;
+        if (null === $date) {
+            return null;
+        }
+
+        // Presumably, this can return false on older PHP versions?
         $parseDate = (array) date_parse($date);
 
         // If no year has been found during date_parse, we nuke the whole value
         // because the date is invalid
-        if (null !== $date && false === $parseDate['year']) {
+        $parsedYear = $parseDate['year'] ?? false;
+        if (false === $parsedYear) {
             $this->logger->info('Date is bad (wrong year): {date}', ['date' => $date]);
 
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


src/Extractor/ContentExtractor.php
```diff
@@ -685,31 +685,39 @@ public function getNextPageUrl(): ?string
...
+            return null;
+        }
+
+        // Presumably, this can return false on older PHP versions?
         $parseDate = (array) date_parse($date);
 
         // If no year has been found during date_parse, we nuke the whole value
         // because the date is invalid
-        if (null !== $date && false === $parseDate['year']) {
+        $parsedYear = $parseDate['year'] ?? false;
```
jtojnar, line 701: I added it for when `date_parse` returns `false` and is cast to array. But it is not clear if that can actually happen. I did not manage to trigger it when I [tried passing it various invalid values](https://3v4l.org/6CQiI). So perhaps the [docs](https://www.php.net/manual/en/function.date-parse.php) are incorrect and it can only return `array` and the cast is pointless.

src/Extractor/ContentExtractor.php
```diff
@@ -685,31 +685,39 @@ public function getNextPageUrl(): ?string
...
 
             return null;
         }
 
-        return (new \DateTime($date))->format(\DateTime::W3C);
+        try {
+            return (new \DateTime($date))->format(\DateTime::W3C);
+        } catch (\Exception $exception) {
+            return null;
+        }
```
jtojnar, line 720: Maybe we should just scrap this hole function in favour of just this `try`-`catch` block.

src/Extractor/ContentExtractor.php
j0k3r: I never got a chance to got a `false` either but it was reported by PHPStan.

src/Extractor/ContentExtractor.php
j0k3r: That's a good idea!
Give it a try.
If it works we just need to move the log inside the catch.

jtojnar:
Closing in favour of #317

PR #316, opened 2023-03-30
Graby: Return object instead of an array

jtojnar:


src/Content.php
```diff
@@ -0,0 +1,140 @@
+<?php
+
+namespace Graby;
+
+/**
+ * Immutable container for Graby result.
+ */
+final class Content
+{
+    private int $status;
```
jtojnar, line 10: Maybe instead of HTTP properties, we could just store PSR-7 response object.

src/Content.php
```diff
@@ -0,0 +1,167 @@
...
+    private string $html;
+
+    private string $title;
+
+    private ?string $language;
+
+    private ?string $date;
+
+    /** @var string[]|null */
+    private ?array $authors;
```
jtojnar, line 21: Looks like author array can contain `false`:

https://github.com/j0k3r/graby/blob/7de23d400791c5c0e343d26475e483519da03338/src/Extractor/ContentExtractor.php#L1338

src/Content.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public function getHtml(): string
+    {
+        return $this->html;
+    }
+
+    public function withHtml(?string $html): self
+    {
+        $new = clone $this;
```
j0k3r, line 76: I'm wondering if this won't use too much ram because of intense cloning during the process (with all these `with*` methods used)?

src/Graby.php
```diff
@@ -412,9 +408,9 @@ private function doFetchContent(string $url): array
             return $res;
         }
 
-        $res['html'] = $this->cleanupHtml($contentBlock, $effectiveUrl);
+        $res = $res->withHtml($this->cleanupHtml($contentBlock, $effectiveUrl));
 
-        $this->logger->info('Returning data (most interesting ones): {data}', ['data' => ['html' => '(only length for debug): ' . \strlen($res['html'])] + $res]);
+        $this->logger->info('Returning data (most interesting ones): {data}', ['data' => $res->withHtml('(only length for debug): ' . \strlen($res->getHtml()))]);
```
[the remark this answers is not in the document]
jtojnar: It would dump the whole object, note the ` + $res`.

src/Content.php
jtojnar: I do not think it would, it is just a shallow copy. And during the chaining, the original copies should be GC'd almost immediately, if the JIT compiler does not eliminate the cloning altogether.

src/Graby.php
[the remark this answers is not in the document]
j0k3r: Good catch, didn't see the `+ $res` with the diff!

src/Content.php
j0k3r: As far as I remember we've added the `status` to be able to retrieve it in wallabag (same as for `headers`).
So maybe we can ditch both and keep one `response`?

jtojnar: Yes, though I would want to return that with `Graby\Extractor\HttpClient::fetch()`. And we would still need some mechanism to pass the effective URL. I guess we can decide on that later.

j0k3r (approved):
Good job!

PR #321, opened 2023-04-11
README: Distinguish dev version usage

jtojnar:
The installation instructions will install a stable version but the example is for an unstable one.

Let’s fix the installation code to install the unstable version and point people to the stable README.


j0k3r (approved):
Thanks

PR #323, opened 2023-04-13
Add async support

jtojnar:
Unfortunately, it currently does not work because `HttpFulfilledPromise` is broken (does not allow to be converted to another promise type).


jtojnar:
I guess we could introduce a dependency on a concrete promise library such as https://github.com/reactphp/promise/tree/2.x

j0k3r:
I never worked on async/promise on PHP so I might not be able to properly review it. But will it force people to use async/promise or will they still be able to use the _sync_ way like before?

jtojnar:
We can provide sync api by just calling `wait()` on the promise (that is what `HttpClient::fetch()` does.

jtojnar:
Tests depend on https://github.com/j0k3r/httplug-ssrf-plugin/pull/10

jtojnar:
Will want to finish https://github.com/j0k3r/graby/pull/324 first, since that way conflicts are easier to resolve.

PR #324, opened 2023-04-14
Return EffectiveResponse from HttpClient

jtojnar:
Another step in more type-safe Graby 3.0.

`HttpClient` will return `EffectiveResponse` instead of an `array`, and `Content` will expose that, instead of having status/headers/url directly.
While at it, we are using `UriInterface` internally instead of `string`.


jtojnar:
I have split the primary commit into several smaller independent ones so it is easier to review commit-by-commit.

j0k3r (approved):
Sorry for being late to review it, everything seems ok to me!
Really good job.

PR #327, opened 2023-05-22
Fix back anchor link

j0k3r:
Also, use `Psr18ClientDiscovery` instead `HttpClientDiscovery` (as reported by PHPStan), see https://docs.php-http.org/en/latest/discovery.html#psr-18-client-discovery

Fix https://github.com/wallabag/wallabag/issues/5315


jtojnar:
Would be nice to include the context from the Wallabag issue in the commit message. (At least that it was originally introduced in https://github.com/j0k3r/graby/commit/ad87368dc6eab4c13814f05b072e1a2904c82444 and probably accidentally reverted in a merge commit.)

j0k3r:
I thought linking the wallabag issue in the PR was enough (as it contains the context)

jtojnar:
It is not in the commit message so it means when doing `git blame` or `git bisect`, I would need to open the commit on GitHub, then open the associated PR, then open Wallabag issue, and then click the commit. And that is assuming GitHub will still be there in a few years.

jtojnar (approved):
LGTM, thanks.

Issue #331, opened 2023-06-23
Using Graby with Laravel


kolaente:
Trying to install Graby in a Laravel application fails because of different monolog requirements:

```
$ composer require 'j0k3r/graby dev-master' php-http/guzzle7-adapter
./composer.json has been updated
Running composer update j0k3r/graby php-http/guzzle7-adapter
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires j0k3r/graby dev-master -> satisfiable by j0k3r/graby[dev-master].
    - j0k3r/graby dev-master requires monolog/monolog ^1.18.0|^2.3 -> found monolog/monolog[1.18.0, ..., 1.27.1, 2.3.0, ..., 2.9.1] but the package is fixed to 3.4.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
You can also try re-running composer require with an explicit version constraint, e.g. "composer require php-http/guzzle7-adapter:*" to figure out if any version is installable, or "composer require php-http/guzzle7-adapter:^2.1" if you know which you need.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

(Trying the 2.x release fails because the guzzle versions are not compatible)

Laravel appears to require at least monolog 3:

```
$ composer why monolog/monolog
laravel/framework v10.13.5 requires monolog/monolog (^3.0)
```

Any way to solve this? Or [do I need to wait for 3.0](https://github.com/j0k3r/graby/issues/298#issuecomment-1269141850)?

kolaente:
Opened a PR: https://github.com/j0k3r/graby/pull/332

PR #332, opened 2023-06-23
feat: upgrade monolog/monolog to v3

kolaente:
Resolves https://github.com/j0k3r/graby/issues/331

This probably needs some more fixing. I've added the nix flake to be able to run composer et al, can remove it if you don't need it.

I had to upgrade `j0k3r/php-readability` to `dev-master` because monolog now requires `psr/log` in at least version 3 and the last release only supports v1. As such, this is probably not mergable right now because we now have a dependency on `master` from that package which will give users something different every time. If you release a new version of that package, I'll happily pin it to that release version.

As per [monolog's release docs](https://github.com/Seldaek/monolog/blob/3.2.0/UPGRADE.md#300), the minimum required php version is now 8.1 so I upgraded that as well. I _think_ I migrated all uses of it but I'm not 100% sure. This probably needs more testing from someone who knows how that feature works.


kolaente:
I guess the tests need an update because we now need 8.1. Should I change them? This probably has other implications I'm not aware of.

jtojnar:
I am afraid raising minimum PHP version at this time is not in line with for Graby primary consumers. Both Wallabag and selfoss still target PHP 7.4. And while I am open to potentially raising it in selfoss, I would still want to support PHP 8.0 at least until it no longer has [upstream security support](https://endoflife.date/php).

src/Monolog/Handler/GrabyHandler.php
```diff
@@ -45,9 +46,9 @@ public function hasRecords($level): bool
         return isset($this->recordsByLevel[$level]);
     }
 
-    protected function write(array $record): void
+    protected function write(LogRecord $record): void
```
jtojnar, line 49: Looks like LogRecord is present in monolog 2.4+ https://github.com/Seldaek/monolog/blob/2.9.1/src/Monolog/LogRecord.php

kolaente:
Yeah that makes sense. Given that php 8.0 will be eol in November, should I pick this up again at that time?

src/Monolog/Handler/GrabyHandler.php
kolaente: So we could do a "soft upgrade", so that the actual upgrade to monolog 3 will only be switching the dependency in composer?

jtojnar: No, it is only available as a stub. The implementation would require accepting `array|LogRecord` for the time being.

jtojnar:
If you are willing to implement it, supporting Monolog `^2.4||^3.0` in master is probably the best solution to make Graby usable with project that require 3.0 now.

I would not bet on dropping PHP 8.0 support in November. There are still distros where PHP 8.0 or even 7.4 will be supported for the next while. For example, [Debian 11](https://endoflife.date/debian), which ships [PHP 7.4](https://wiki.debian.org/PHP), will continue to receive security updates for three more years.

kolaente:
> If you are willing to implement it, supporting Monolog ^2.4||^3.0 in master is probably the best solution to make Graby usable with project that require 3.0 now.

Done

kolaente:
Now the tests fail with the same issue as in https://github.com/j0k3r/graby/issues/333

composer.json
```diff
@@ -15,15 +15,15 @@
...
+        "php": ">=7.4|>=8.1",
         "ext-curl": "*",
         "ext-tidy": "*",
         "fossar/htmlawed": "^1.2.8",
         "guzzlehttp/psr7": "^2.0",
         "j0k3r/graby-site-config": "^1.0.147",
         "j0k3r/httplug-ssrf-plugin": "^2.0",
-        "j0k3r/php-readability": "^1.2.9",
-        "monolog/monolog": "^1.18.0|^2.3",
+        "j0k3r/php-readability": "dev-master",
```
jtojnar, line 25: Is there any reason you are bumping readability?

j0k3r: Because of `psr/log`, it's detailed in the PR description

kolaente: Because readability 1x only supports psr/log 1.

kolaente: Should I try 2.0.2?

j0k3r: Maybe you can put `^2.0` it'll be enough for the `psr/log` v3

flake.nix
```diff
@@ -0,0 +1,44 @@
...
+  description = "Dev environment";
+
+  outputs = { self, nixpkgs }:
+    let pkgs = nixpkgs.legacyPackages.x86_64-linux;
+    in {
+      defaultPackage.x86_64-linux =
+        pkgs.mkShell {
+          buildInputs =  with pkgs; [
+            (php82.withExtensions
+              ({ all, ... }: with all; [
```
jtojnar, line 11: Not sure if @j0k3r wants to include Nix packaging. I could continue to maintain it for the time being as I use something like this locally.

----

It might be sufficient to use `enabled` attribute to load the default extensions:

https://github.com/fossar/selfoss/blob/b0eb43d80bedde5da3204b842511afe245d8c17c/flake.nix#L37-L40

And make the PHP version parametric:

https://github.com/fossar/guzzle-transcoder/blob/816585c89c87268d1188d2d9696b49ce73541ff9/flake.nix#L29

flake.nix
```diff
@@ -0,0 +1,44 @@
...
+                xmlwriter
+#                yaml
+                zip
+                zlib
+              ])
+            )
+            php82Packages.composer
+            libjpeg
+            libpng
+            libyaml
```
jtojnar, line 41: What are these used for?

kolaente: I've copied the flake from my own project where I need those - here they're not needed. I removed them.

flake.nix
j0k3r: If you can maintain it, I'm ok.
I didn't what Flake was.
Maybe put a comment at the very top of `flake.nix` with a link to sth that explain what it does/is.

composer.json
```diff
@@ -15,15 +15,15 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php": ">=7.4",
+        "php": ">=7.4|>=8.1",
```
jtojnar, line 18: This is superfluous.

src/Monolog/Formatter/GrabyFormatter.php
```diff
@@ -61,18 +67,18 @@ protected function convertToString($data): string
...
      * Creates an HTML table row with background cellon title cell.
      *
-     * @param int    $level    Error level
+     * @param Level  $level    Error level
      * @param string $th       Row header content
      * @param string $td       Row standard cell content
      * @param bool   $escapeTd false if td content must not be html escaped
      */
-    private function addRowWithLevel(int $level, string $th, string $td = ' ', bool $escapeTd = true): string
+    private function addRowWithLevel(Level $level, string $th, string $td = ' ', bool $escapeTd = true): string
```
jtojnar, line 75: This will not work with Monolog 2, will it?

tacman: Hmm, I'm getting this error, I think I have monolog 2 installed.

    Graby\Monolog\Formatter\GrabyFormatter::addRowWithLevel(): Argument #2 ($th) must be of type string, int gi

PR #336, opened 2023-08-12
Improve regex used to remove conditional comments in HTML

Kdecherf:
From the test added by this commit, the previous regex incorrectly removed the shown.jpg image. The new regex now keeps what would be normally shown by a browser.

It should be noted that, according to regex101.com this instruction comes at a cost three times higher that before (~1,200 iterations in place of ~400)

Fixes https://github.com/wallabag/wallabag/issues/6828

**This must be backported into release/2.x for wallabag**

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -179,27 +179,9 @@ class HttpClient
 
         $body = (string) $response->getBody();
 
-        // be sure to remove conditional comments for IE around the html tag
-        // we only remove conditional comments until we found the <head> tag
-        // they usually contains the <html> tag which we try to found and replace the last occurence
-        // with the whole conditional comments
-        preg_match('/^\<!--\[if(\X+)\<!\[endif\]--\>(\X+)\<head\>$/mi', $body, $matchesConditional);
-
-        if (\count($matchesConditional) > 1) {
-            preg_match_all('/\<html([\sa-z0-9\=\"\"\-:\/\.\#]+)\>$/mi', $matchesConditional[0], $matchesHtml);
-
-            if (\count($matchesHtml) > 1) {
-                $htmlTag = end($matchesHtml[0]);
-
-                if (!empty($htmlTag)) {
-                    $body = str_replace($matchesConditional[0], $htmlTag . '<head>', $body);
-                }
-            }
-        }
-
         // be sure to remove ALL other conditional comments for IE
-        // (regex found here: https://stackoverflow.com/a/137831/569101)
-        preg_match_all('/<!--\[if\s(?:[^<]+|<(?!!\[endif\]-->))*<!\[endif\]-->/mi', $body, $matchesConditional);
+        // (regex inspired from here: https://stackoverflow.com/a/55083809/954513)
+        preg_match_all('/<!--(?:\[| ?<!).+?-->/mis', $body, $matchesConditional);
 
         if (isset($matchesConditional[0]) && (is_countable($matchesConditional[0]) ? \count($matchesConditional[0]) : 0) > 1) {
             foreach ($matchesConditional as $conditionalComment) {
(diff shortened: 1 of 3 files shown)


tests/Extractor/HttpClientTest.php
```diff
@@ -372,13 +372,29 @@ public function dataForConditionalComments(): array
...
+                'presentData' => 'shown image',
+            ],
         ];
     }
 
     /**
      * @dataProvider dataForConditionalComments
      */
-    public function testWithMetaRefreshInConditionalComments(string $url, string $html, string $removeData): void
+    public function testWithMetaRefreshInConditionalComments(string $url, string $html, string $removeData, ?string $presentData = null): void
```
jtojnar, line 397: Maybe use `keepData`, `preserveData` or `retainData` to make it clearer that it should be present after the extraction.

Kdecherf: updated with `preserveData`

src/Extractor/HttpClient.php
```diff
@@ -198,8 +198,8 @@ public function fetch(UriInterface $url, bool $skipTypeVerification = false, arr
         }
```
jtojnar, line 198: The whole heuristics above is prety ugly. What if there is no `<head>` tag? What if the `<html>` tag contains an attribute that includes a character not listed above. I wonder if it were not cleaner to just remove all downlevel-hidden conditional comments including their contents, and then remove all downlevel-revealed comments. Or just remove all HTML comments completely.

Kdecherf: Removing all HTML comments would probably be easier

src/Extractor/HttpClient.php
```diff
@@ -198,8 +198,8 @@ public function fetch(UriInterface $url, bool $skipTypeVerification = false, arr
...
 
         // be sure to remove ALL other conditional comments for IE
-        // (regex found here: https://stackoverflow.com/a/137831/569101)
-        preg_match_all('/<!--\[if\s(?:[^<]+|<(?!!\[endif\]-->))*<!\[endif\]-->/mi', $body, $matchesConditional);
+        // (regex inspired from here: https://stackoverflow.com/a/55083809/954513)
+        preg_match_all('/<!--(?:\[| ?<!)(?:[^-]|-[^-]|--+[^->])*--+>/mi', $body, $matchesConditional);
```
jtojnar, line 202: The regex looks mostly correct to me:

- The left branch in the first non-capturing group will match [downlevel-hidden conditional comments](https://en.wikipedia.org/wiki/Conditional_comment#Downlevel-hidden_conditional_comment) and the rest of the regex will match remainder of the conditional. It will also match just the opening part of the [downlevel-revealed tag](https://en.wikipedia.org/wiki/Conditional_comment#Downlevel-revealed_conditional_comment) encapsulated within a comment.
- The right branch will then match the closing part of the downlevel-revealed tag (encapsulated within a comment).
- Comments [cannot really contain `--` so](https://developer.mozilla.org/en-US/docs/Web/API/Comment) the `--+[^->]` branch is redundant.

It will not handle standalone downlevel-revealed tags but those were not handled before either and are probably rare since they are not valid HTML syntax.

src/Extractor/HttpClient.php
jtojnar: Looking at the new regex, it should do the former, so the HTML stuff above should be redundant.

Kdecherf: done e13c8e952a06b2ea13db07fca2668b48c4afd95d

Kdecherf: Done e13c8e952a06b2ea13db07fca2668b48c4afd95d

src/Extractor/HttpClient.php
Kdecherf: Removing `|--+[^->` actually breaks the original case I am fixing as it doesn't match the `if [!mso]` line

2023-08-13-195136_736x279

jtojnar: Ah. That is actually not a valid HTML – it should be `<!--[if !mso]><!-->`. But I guess if real sites use this, we need to support this abomination as well. Maybe add a comment that that branch is to support non-compliant

tests/Extractor/HttpClientTest.php
```diff
@@ -372,13 +372,29 @@ public function dataForConditionalComments(): array
...
                 'removeData' => '<meta http-equiv="refresh" content="1;url=/2.85.0/static/unsupp.html" />',
             ],
+            [
```
jtojnar, line 375: Labelling the instance would be nice too:

```suggestion
            'non-standard downlevel-revealed comment' => [
```

That way, we will be able to refer to it in the implementation.

src/Extractor/HttpClient.php
jtojnar: Actually, looks like [HTML 5 now allows double dashes in comments](https://github.com/whatwg/html/issues/1351) so maybe just replace `(?:[^-]|-[^-]|--+[^->])*--+>` with a non-greedy match like:

```suggestion
        preg_match_all('/<!--(?:\[| ?<!).+?-->/mi', $body, $matchesConditional);
```

Kdecherf: Your pattern does not match all things matched by mine:
2023-09-12-210344_717x420

tests/Extractor/HttpClientTest.php
Kdecherf: Done e337a82

src/Extractor/HttpClient.php
jtojnar: Oh, right, it will also need `dotall` (s) flag to have `.` match line breaks.

Kdecherf: It looks like to work indeed 8c225bd05edac28ca323070b78a68ccca9afbc76

j0k3r (commented):
Looks great, you can squash :+1:

j0k3r (approved):

PR #338, opened 2023-08-20
Additional fingerprints

HolgerAusB:
* new fingerprint for ippen.media based newspapers (German), the old one is not longer valid
* new fingerprint for medium.com based websites, transformed from @fivefilters FulltextRSS config


fivefilters:
There's also this for Substack that we use, if you'd like to add it, @HolgerAusB:

`'<link rel="stylesheet" type="text/css" href="https://substackcdn.com/'` -> fingerprint.substack.com

j0k3r (approved):

HolgerAusB:
OK @j0k3r, I don't understand, for what ContentExtractorTest.php is. No need to explain, I am not a real coder!

But you use here a wider range of the fingerprint and I just want be sure that this will not cause problems. You wrote:
`... {"de.ippen-digital.story.onlineId":91197383} ...`
but that number should be different for each article, I think.

And there is not test entry for the other two services I added: fingerprint.medium.com and fingerprint.substack.com

HolgerAusB:
@j0k3r, any ETA about when this is might going to be released and pushed to wallabag?

PR #341, opened 2024-01-03
[2.x] Update deps

j0k3r:
- update GitHub Actions to latest version
- run PHP-CS-Fixer
- use `Psr\Http\Client\ClientInterface` instead of `Http\Client\HttpClient`
- use `Http\Discovery\Psr18ClientDiscovery` instead of `Http\Discovery\HttpClientDiscovery`
- remove the only `utf8_encode` used (because it's deprecated in PHP 8.2) and I don't think these lines are still used (the associated test was already removed)


Kdecherf (commented):
On the removal of `utf8_encode`, did you check that it does not break https://github.com/wallabag/wallabag/issues/2387?

j0k3r:
@Kdecherf yep I've checked, nothing is broken

src/Extractor/HttpClient.php
```diff
@@ -42,11 +42,7 @@ class HttpClient
      */
     private $responseHistory;
 
-    /**
-     * @param Client $client Http client
-     * @param array  $config
-     */
-    public function __construct(Client $client, $config = [], LoggerInterface $logger = null)
+    public function __construct(ClientInterface $client, $config = [], LoggerInterface $logger = null)
```
jtojnar, line 45: `Client` → `ClientInterface`  should be fine. `Client` has been a subtype since 2.0 (https://github.com/php-http/httplug/commit/6885b6c94751e64efc1473bedff1706cf86596d9).

Kdecherf (approved):

PR #343, opened 2024-03-15
Allow monolog/monolog 3

tacman:


j0k3r:
It's not that easy.
And a work is in progress (or at least started) over here: https://github.com/j0k3r/graby/pull/332

tacman:
Oh, drat.  Is there a way around this, so it can be installed applications that use monolog 3, even if logging doesn't work correctly?  

This (and site-config) are very cool libraries, thanks for releasing them!  The application where I want to use them is Symfony 7 and PHP 8.3.  I'll watch #332.

PR #345, opened 2024-03-15
Add default values for parameters after optional parameters

### LINKED ISSUE #344: Optional Parameters can only come after required in PHP 8
Using php 8.3 throws these deprecations:

Deprecated: Optional parameter $node declared before required parameter $logMessage is implicitly treated as a required parameter in /home/tac/g/tacman/f43.me/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php on line 890

Deprecated: Optional parameter $node declared before required parameter $logMessage is implicitly treated as a required parameter in /home/tac/g/tacman/f43.me/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php on line 914

Deprecated: Optional parameter $node declared before required parameter $type is implicitly treated as a required parameter in /home/tac/g/tacman/f43.me/vendor/j0k3r/graby/src/Extractor/ContentExtractor.php on line 985

tacman:
Fixes #344

Not sure that $type can be null, but this at least gets rid of the deprecation errors.

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -887,7 +887,7 @@ class ContentExtractor
      *
      * @return bool Telling if we have to detect title again or not
      */
-    private function extractTitle(bool $detectTitle, string $cssClass, ?\DOMNode $node = null, string $logMessage): bool
+    private function extractTitle(bool $detectTitle, string $cssClass, ?\DOMNode $node = null, ?string $logMessage=null): bool
     {
         if (null === $node) {
             return true;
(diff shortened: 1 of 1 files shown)


jtojnar (changes requested):
`$logMessage` and `$type` should not become nullable since they are concatenated to strings.

AFAICT, the `$node` arguments are always passed so we should remove the default there.

tacman:
Makes sense.  Should I close this PR and submit a new one with $node as a required parameter?

jtojnar:
Feel free to force push here

jtojnar:
Superseded by https://github.com/j0k3r/graby/pull/347

PR #346, opened 2024-03-15
Add php 8 to matrix

tacman:


j0k3r:
It's already there but not required.
image

PR #347, opened 2024-03-15
ContentExtractor: Make $node a required parameter

### LINKED ISSUE #344: Optional Parameters can only come after required in PHP 8

tacman:
Fixes #344

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -887,7 +887,7 @@ class ContentExtractor
      *
      * @return bool Telling if we have to detect title again or not
      */
-    private function extractTitle(bool $detectTitle, string $cssClass, ?\DOMNode $node = null, string $logMessage): bool
+    private function extractTitle(bool $detectTitle, string $cssClass, ?\DOMNode $node, string $logMessage): bool
     {
         if (null === $node) {
             return true;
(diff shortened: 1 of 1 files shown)


jtojnar:
Could you please also update the other methods like `extractAuthor` for consistency?

tacman:
My logic was to do as little as possible to fix the error.  And unfortunately, I'm not that good with PRs like this!  Can you accept these, then I'll do a new PR?

And I guess I should close my other ones, although I don't know how this passes PHP 8.3 which is why I added it to the matrix.

jtojnar:
Yes, generally, a focused patch is best, though it should not break other aspects of the code, be it formal ones (e.g. coding style) or informal (e.g. consistency).

I have added a fix for the `extractAuthor` case.

Thanks for your contribution.

PR #352, opened 2024-09-27
Keep empty img tags

Strubbl:
keep empty img tags to avoid missing images, e.g. https://github.com/wallabag/wallabag/issues/7700

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -319,8 +319,8 @@ class Graby
         $re = '/^[ \t]*[\r\n]+/m';
         $htmlCleaned = preg_replace($re, '', $html);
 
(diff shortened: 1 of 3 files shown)


jtojnar:
Thanks for the contribution.

Would you be able to include a test like the following?

https://github.com/j0k3r/graby/blob/1281bf3d7045d2f2682d1af6ba3715e492184e9a/tests/GrabyTest.php#L1000-L1008

----

Aside, I find it weird that https://github.com/j0k3r/graby/pull/141 was made against Graby rather than Readability in the first place but whatever 🤷‍♀️

HolgerAusB:
Thank you @Strubbl and @j0k3r, that will help a lot. 💕💕💕

PR #353, opened 2024-09-28
Fix CI

jtojnar:
- Fix coding style with newer PHP-CS-Fixer
- Fix analysis with newer PHPStan
- Remove testing against Guzzle 5

PR #354, opened 2024-09-29
PHPStan fixes

jtojnar:
Now only the `JSLikeHTMLElement` issues remain.

Depends on https://github.com/j0k3r/graby/pull/353


j0k3r (approved):
Really good job!

PR #286, opened 2022-02-04
Update codebase to support Readability 2

Kdecherf:

PR #356, opened 2025-01-13
Update version j0k3r/php-readability

Grandroot:
Upgrade j0k3r/php-readability version

It avoids the following error on project creation : 
`Fatal error: Declaration of Readability\Readability::setLogger(Psr\Log\LoggerInterface $logger) must be compatible with Psr\Log\LoggerAwareInterface::setLogger(Psr\Log\LoggerInterface $logger): void`

Why ? In php-readability version 1.2.9 (or even 1.2.10) the function setLogger wasn't typed void, unlike Psr\Log\LoggerAwareInterface::setLogger, and that's the cause of the error

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "guzzlehttp/psr7": "^2.0",
         "j0k3r/graby-site-config": "^1.0.181",
         "j0k3r/httplug-ssrf-plugin": "^2.0",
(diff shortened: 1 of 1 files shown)


j0k3r:
Work started here instead https://github.com/j0k3r/graby/pull/286

Issue #359, opened 2025-02-09
Unproper handling of National Characters


ArturRuta:
**Note:** This originally arised as an issue in wallabag but initial research showed that it's very likely coming from graby which wallabag uses.

# The problem is
on oncasion the national characters in an article text are not shown as expected ans instead they are replaced by extrange combinations of garbage characters.

The issue is not arising for all articles, some correctly show the natinoal characters while other doesn't. 
The behavior is consistent in the sense that for a guiven article the result is allways good or allways bad.
Curiously enough the article's title is allways properly handled (at least in the samples I tried) while the content may or may not be properly transcribed.
- One sample of a bad behavior comes from this url: [error example](https://www.xataka.com/movilidad/coches-vendidos-2023-2024-espana) 
  - You will find that the article tittle, even if containing national characters it's properly handled. For example it contais the work: **más**
  - On the other side, the article contents are not propoerly handled. Very early in the text you can see for example the word **automÃ³vil** that is wrong. It should look like **automóvil** instead.
- One sample of proper behavior comes from this url: [correct sample](https://www.xataka.com/movilidad/byd-sealion-7-opiniones-primera-toma-contacto-fotos)

# When digging into this topic i saw the following:
- Wallabag stores the article's content with the wrong characters, hence the problem is not how the result is presented. It lays in the way the source is processed.
- Another application, f43.me which also relies on graby shows the same kind of misbehavior
- Finally, i managed to enable graby logs in wallabag (se attached content), and while I have a limited understanding it seems that the garbage characters show up there.

I cannot do stantalone tests of graby, that falls well behing my capabilities, but if the issues is found/fixes i would be very glad of testing any wallabag o f43.me releases includig the fix.

By the way, wallabag version is  2.6.10 which is the latest one at the time of this writing, but i cannot found which graby version is packed inside.

Any help will be greatly appreciated
Best regards.

[graby.log](https://github.com/user-attachments/files/18725443/graby.log)

ArturRuta:
Unfortunately this was a deal breaker for me.
Since no fix was available I had to move to hoarder which hasn't this issue.
Anyway I'm open to help testing/verifying a solution if/whenever is developped

jtojnar:
Thanks for reporting and sorry for the late reply. I am able to reproduce it on master with the following:

```php
<?php

declare(strict_types=1);

require_once __DIR__ . '/vendor/autoload.php';

use Monolog\Formatter\LineFormatter;
[...]

PR #366, opened 2025-02-23
ContentExtractor: Fix encoding in JSON-LD

jtojnar:
When page has an article body specified in JSON-LD, the HTML fragment is passed to `DOMDocument::loadHTML`.
`DOMDocument` will parse a HTML document as ISO-8859-1, unless the document contains an XML encoding declaration or HTML meta tag setting character set.

https://stackoverflow.com/a/39148511/160386

JSON-LD typically does not contain `meta[charset]` tag so we need to add it ourselves.

At this point the input should already be in UTF-8, as we convert all inputs to that encoding everywhere except in `Graby::cleanupHtml`. There we will clarify the requirement in doc comment.

Fixes: https://github.com/j0k3r/graby/issues/359

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -1208,7 +1208,7 @@ class ContentExtractor
      *     - OpenGraph
      *     - JSON-LD.
      *
-     * @param string $html Html from the page
+     * @param string $html UTF-8-encoded HTML fragment of the page
      */
     private function extractDefinedInformation(string $html): void
     {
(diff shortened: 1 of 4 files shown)


j0k3r (approved):

ArturRuta:
I will notify the wallabag group so that eventually they can get the new version of graby and hopefully we will see it working properly.
Thanks a lot for helping

j0k3r:
@ArturRuta the new version of Graby will be updated in the coming hours I guess and it'll be available in wallabag 2.6.11 (no ETA yet)

Issue #275, opened 2021-12-10
Graby 3.0


j0k3r:
I'm thinking about the 3.0.
Not to much breaking changes but mostly type everything (method args, return, variable, etc.) for now and dropping PHP < 7.4

See #272 & #273

I would like to know if you think about something that should be broken in v2 or re-implemented in a way that will require a major version.
So it can be included in 3.0.
Of course, "rewrite everything because method XXX sucks" isn't an appropriate answer.

If this is something really too big and might take too much time to complete, it can also be part of v4 :)

j0k3r:
I've also removed `http-interop/http-factory-guzzle` because `guzzle/psr7` 2.0 finally have a PSR-17 factory but it means v3 will drop support for Guzzle 6 because v6 is locked to `guzzle/psr7` v1. This does not affect Guzzle 5

See https://github.com/j0k3r/graby/pull/277

nicosomb:
Do we need to wait for graby 3.0 if we want to start a v3 POC for wallabag?

j0k3r:
Not really, we can use the master.

Kdecherf:
Was wondering what we could do with our guzzle-site-authenticator plugin, if it should be done differently or not

jtojnar:
My wishlist:

- [ ] Removing the use of arrays from return values #316
- [ ] Having constructors take config objects instead of arrays.
- [ ] Making `ContentExtractor` stateless – so that there is no need to call `reset()`.
- [ ] Making `ContentExtractor::$readability` property private (possibly constructed by `Graby` and injected)
- [ ] Possibly using PSR-18 directly instead of HTTPPlug – but there is no support for plug-ins so not sure if it is feasible.
- [ ] Passing PHPStan level `max`

j0k3r:
> Having constructors take config objects instead of arrays.

We already have Value Object (https://github.com/j0k3r/graby/pull/284) so won't be to hard to move to object as parameter

> Making `ContentExtractor::$readability` property private

How can we define configuration for readability when using Graby from an external source then?

jtojnar:
I would also like to see

- [ ] async/promise support

It would mean switching from [psr/http-client-implementation](https://packagist.org/providers/psr/http-client-implementation) to [php-http/async-client-implementation](https://packagist.org/providers/php-http/async-client-implementation) and using `Http\Promise\Promise` in the API.

tacman:
[...]

PR #378, opened 2025-09-26
Making ContentExtractor stateless

j0k3r:
Started with Claude to have a first draft and then update the whole stuff.

I'm just wondering if `ExtractedContent` should have getters or all keys public instead.

See https://github.com/j0k3r/graby/issues/275#issuecomment-1491142099


j0k3r:
Waiting for https://github.com/j0k3r/graby/pull/379

jtojnar:
> all keys public instead.

With PHP 8.1 `readonly` properties, public properties sound good enough.

PR #379, opened 2025-09-26
Remove PHP < 8.2

j0k3r:


.github/workflows/continuous-integration.yml
jtojnar, line 1: Looks like upstream PHP 8.1 security support is [ending in three months](https://www.php.net/supported-versions.php). Though, [Debian 11 bullseye](https://wiki.debian.org/PHP#PHP_version_by_Debian_release), which ships PHP 7.4, will continue to be supported for [one more year](https://wiki.debian.org/DebianReleases#Versions_of_Debian_stable). And while I would typically follow Debian, the improvements in PHP 8 do sound quite tempting.

maintenance/Rector/MockGrabyResponseRector.php
```diff
@@ -42,21 +42,9 @@ final class MockGrabyResponseRector extends AbstractRector
...
-        UseNodesToAddCollector $useNodesToAddCollector,
-        ValueResolver $valueResolver,
-        VariableNaming $variableNaming
-    ) {
-        $this->betterNodeFinder = $betterNodeFinder;
-        $this->useNodesToAddCollector = $useNodesToAddCollector;
-        $this->valueResolver = $valueResolver;
-        $this->variableNaming = $variableNaming;
+
+    public function __construct(private readonly BetterNodeFinder $betterNodeFinder, private readonly UseNodesToAddCollector $useNodesToAddCollector, private readonly ValueResolver $valueResolver, private readonly VariableNaming $variableNaming)
```
jtojnar, line 46: Would be more readable formatted like this (https://cs.symfony.com/doc/rules/function_notation/method_argument_space.html)


```suggestion
    public function __construct(
        private readonly BetterNodeFinder $betterNodeFinder,
        private readonly UseNodesToAddCollector $useNodesToAddCollector,
        private readonly ValueResolver $valueResolver,
        private readonly VariableNaming $variableNaming,
    )
```

src/GrabyConfig.php
jtojnar, line 1: We can probably drop the options resolver and just use PHP 8 named arguments in a constructor. But that should go to later PR.

maintenance/Rector/MockGrabyResponseRector.php
```diff
@@ -42,21 +42,9 @@ final class MockGrabyResponseRector extends AbstractRector
     private const FIXTURE_DIRECTORY = __DIR__ . '/../../tests/fixtures/content';
     private const MATCHING_ERROR_COMMENT = 'TODO: Rector was unable to evaluate this Graby config.';
     private const IGNORE_COMMENT = 'Rector: do not add mock client';
-    private BetterNodeFinder $betterNodeFinder;
```
jtojnar, line 45: I assume these changes are made by rector but have no bearing on ability to run the code. It would be easier to review if they were split into a separate commit.

j0k3r: Yeah it was done by rector as `maintenance` folder is part of the config.

PR #381, opened 2025-12-12
Update regex for empty nodes for also excluding img

HolgerAusB:
Some websites using closing `</img>` nodes which is not defined in HTML. In most cases, this results in `</img>` which is currently detected as empty node and therefore removed by graby. So all images from those domains can currently not be fetched by wallabag.

This will fix lot of issues in wallabagger for missing images.

example: https://taz.de/Stadtbilddebatte-wird-praktisch/!6136666/


j0k3r:
You are right, this is a duplicate.

Duplicate of #352

PR #387, opened 2026-05-25
perf: improve performance of absolute link generation on large pages

deviantintegral:
I noticed that some (long) Hacker News pages were taking over 30 seconds to generate and sometimes timing out in my configuration. It turned out most of the time was spent in `makeAbsolute`.

This PR does two things:

- First adds a bit of missing test coverage around the child methods.
- Then brings in the actual performance fix which only adds tests, so hopefully there's no hidden regressions here!

Locally (on much faster hardware than where I was getting timeouts) I get:

  | n (elements) | old (ms) | new (ms) | speedup |
  |---:|---:|---:|---:|
  | 900 | 6.2 | 0.7 | 8.8× |
  | 4,500 | 149 | 3.4 | 43× |
  | 9,000 | 627 | 6.8 | 92× |
  | 18,000 | 3,466 | 10.0 | 346× |
  | 30,000 | 10,637 | 17.2 | 618× |

This is AI-assisted. I'm glad to answer any questions or make changes.

### DIFF
--- a/src/Graby.php
+++ b/src/Graby.php
@@ -798,18 +798,16 @@ class Graby
      */
     private function makeAbsolute(UriInterface $base, \DOMElement $elem): void
     {
-        foreach (['a' => 'href', 'img' => 'src', 'iframe' => 'src'] as $tag => $attr) {
-            $elems = $elem->getElementsByTagName($tag);
+        $tagAttrMap = ['a' => 'href', 'img' => 'src', 'iframe' => 'src'];
 
-            for ($i = $elems->length - 1; $i >= 0; --$i) {
-                $e = $elems->item($i);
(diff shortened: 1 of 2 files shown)


jtojnar:
Could you please outline what the problem was and how you fixed it?

**Edit:** Never mind, I see it in the comment. Could you please move it into the commit message? Those comments do not make sense in the code without the previous code.

deviantintegral:
Good thought, done! I think the previous CI error in https://github.com/j0k3r/graby/actions/runs/26380608738/job/77805518411 is unrelated to the PR so I think this will likely pass when the jobs are approved.

j0k3r (approved):
Thanks!

PR #396, opened 2026-07-08
Handle cookie-gate redirects that loop to the same URL

beginner59:
## Summary

Some sites (for example Tamedia/Piano CMS pages such as `tagesanzeiger.ch`) respond with:

- `302 Found`
- `Set-Cookie`
- `Location` pointing back to the same URL

In this case `RedirectPlugin` raises `CircularRedirectionException` before the cookie plugin can persist the cookie, so the follow-up request never includes the entitlement token.

This patch fixes that by:

- adding `CookiePlugin` to the plugin chain
- exposing `CookiePlugin::storeSetCookies()` for manual persistence
- detecting this specific cookie-gate redirect pattern
- storing cookies from the redirect response
- retrying the request once with the stored cookie

## Test plan

- Added `testCookieGateRedirectRetriesWithStoredCookie`
- Verified that the retry request includes the stored `entitlementToken`

### DIFF
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -274,10 +274,6 @@ class HttpClient
      */
     private function isCookieGateRedirect(CircularRedirectionException $e): bool
     {
-        if (!method_exists($e, 'getRequest') || !method_exists($e, 'getResponse')) {
-            return false;
-        }
-
         $request = $e->getRequest();
         $response = $e->getResponse();
 
--- a/src/Extractor/HttpClient.php
+++ b/src/Extractor/HttpClient.php
@@ -5,9 +5,12 @@ declare(strict_types=1);
 namespace Graby\Extractor;
 
 use Graby\HttpClient\EffectiveResponse;
+use Graby\HttpClient\Plugin\CookiePlugin;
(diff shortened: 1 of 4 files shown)


src/Extractor/HttpClient.php
```diff
@@ -69,12 +75,16 @@ public function __construct(
...
+        $this->cookieJar = new CookieJar();
+        $this->cookiePlugin = new CookiePlugin($this->cookieJar);
         $this->client = new HttpMethodsClient(
             new PluginClient(
                 $client,
                 [
                     new ServerSideRequestForgeryProtectionPlugin(),
                     new RedirectPlugin(),
+                    // Inner plugin: stores Set-Cookie before RedirectPlugin handles the response.
+                    $this->cookiePlugin,
```
j0k3r, line 87: As you are injecting the cookie plugin here (and it's already done in Graby.php (line 131)) maybe you should remove the default cookie plugin injection in Graby.php

From:
```php
$client ?: new PluginClient(Psr18ClientDiscovery::find(), [new CookiePlugin(new CookieJar())]),
```

to
```php
$client ?: Psr18ClientDiscovery::find(),
```

PR #397, opened 2026-07-12
ContentExtractor: Resolve PHPStan argument type issues

jtojnar:
`DOMXPath::query()` returns `DOMNodeList` that can contain `DOMNameSpaceNode` items in addition to `DOMNode` items. Now, `DOMNameSpaceNode` items will only be returned when ‘[namespace axis]’ is used, which is unlikely as it is useless for content but PHPStan is not currently smart enough to determine that the axis is not used in the static XPath expression, and there is nothing it could do for dynamic expressions so the inferred return type includes `DOMNameSpaceNode`.

Let’s fix the PHPStan errors by:

1. Adding asserts in cases where the XPath expression is static so we know the item types.
2. Extending helper method types to take the returned value.
3. Adding guards to where the XPath expression is dynamic, though it is improbably anyone will ever use an XPath expression that matches a namespace.
4. Introduce `useFirst` helper in cases where are looking for the first `DomNode`.

[namespace axis]: https://www.w3.org/TR/xpath-10/#axes

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -243,6 +243,8 @@ class ContentExtractor
 
             if ($elems instanceof \DOMNodeList && $elems->length > 0) {
                 foreach ($elems as $elem) {
+                    // For PHPStan: The static XPath expression guarantees the type.
+                    \assert($elem instanceof \DOMAttr);
                     $language = trim($elem->textContent);
                     $this->logger->info('Language matched: {language}', ['language' => $language]);
                 }
@@ -374,9 +376,9 @@ class ContentExtractor
             // check for hentry
             $elems = $xpath->query("//*[contains(concat(' ',normalize-space(@class),' '),' hentry ')]", $readability->dom);
 
-            if (false !== $elems && $this->hasElements($elems)) {
+            $hentry = self::firstNode($elems);
+            if (null !== $hentry) {
                 $this->logger->info('hNews: found hentry');
-                $hentry = $elems->item(0);
 
                 // check for entry-title
                 $extractedTitle = $this->extractTitle(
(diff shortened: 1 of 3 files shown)


src/Extractor/ContentExtractor.php
```diff
@@ -760,23 +762,42 @@ private function prepareSiteConfig(string $html, UriInterface $url, ?SiteConfig
     }
 
     /**
-     * Check if given node list exists and has length more than 0.
+     * Check if given node list exists and has contains at least one `\DOMNode`.
```
```suggestion
     * Check if given node list exists and contains at least one `\DOMNode`.
```

src/Extractor/ContentExtractor.php
```diff
@@ -760,23 +762,42 @@ private function prepareSiteConfig(string $html, UriInterface $url, ?SiteConfig
...
+
+    /**
+     * Check if given node list exists and returns the first `\DOMNode` if present.
+     *
+     * @param \DOMNodeList<\DOMNode|\DOMNameSpaceNode>|false $elems Not force typed because it can also be false
+     */
+    private static function firstNode($elems = false): ?\DOMNode
+    {
+        // Internal XPath expressions are valid and evaluating invalid external ones will return an error before getting here.
+        // @codeCoverageIgnoreStart
```
j0k3r, line 782: Because it's impossible to test?

jtojnar: Yeah, as mentioned in the comment, no matter what we pass to the public methods, this line will never be executed.

src/Extractor/ContentExtractor.php
jtojnar: Fixed.

j0k3r (approved):

PR #398, opened 2026-07-23
Add support for guzzlehttp/psr7 3.0

jtojnar:
Unrestrict consumers’ choice.

There are no backwards incompatible changes relevant to us and version 3.0 still supports PHP 7.4.
But let’s support both for a while so that people can migrate gradually.

https://github.com/guzzle/psr7/releases/tag/3.0.0


j0k3r (approved):

j0k3r:
> It was introduced in 63198ef without a rationale.

I think it was added automatically by composer.

PR #399, opened 2026-07-24
Graby: Use typed config

jtojnar:
More type safety and allows us to get rid of the horrendous type annotations in PHPDoc.

Part of https://github.com/j0k3r/graby/issues/275

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ configure it and inject it to `Graby\Graby`.
     $guzzle = new GuzzleClient([
         'timeout' => 2,
     ]);
-    $graby = new Graby([], new GuzzleAdapter($guzzle));
+    $graby = new Graby(client: new GuzzleAdapter($guzzle));
     ```
 
 
--- a/README.md
+++ b/README.md
@@ -211,45 +211,45 @@ configure it and inject it to `Graby\Graby`.
 This is the full documented configuration and also the default one.
 
 ```php
-$graby = new Graby([
+$graby = new Graby(new GrabyConfig(
     // Enable or disable debugging.
     // This will only generate log information in a file (log/graby.log)
-    'debug' => false,
+    debug: false,
     // use LogLevel::Debug value if you want more data (HTML at each step for example) to be dumped in a different file (log/html.log)
-    'log_level' => LogLevel::Info,
+    logLevel: LogLevel::Info,
     // If enabled relative URLs found in the extracted content are automatically rewritten as absolute URLs.
-    'rewrite_relative_urls' => true,
+    rewriteRelativeUrls: true,
     // If enabled, we will try to follow single page links (e.g. print view) on multi-page articles.
     // Currently this only happens for sites where single_page_link has been defined
     // in a site config file.
-    'singlepage' => true,
+    singlepage: true,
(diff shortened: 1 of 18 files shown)


j0k3r (dismissed):
Good job!

src/Extractor/ContentExtractorConfig.php
```diff
@@ -11,93 +11,68 @@
...
+         * @var array{
+         *   pre_filters: array<string, string>,
+         *   post_filters: array<string, string>,
+         * }
+         */
+        public array $readability = [
+            'pre_filters' => [],
+            'post_filters' => [],
+        ],
+        /** @var array<string> */
```
j0k3r, line 36: Shouldn't we use `readability` from the resolved config instead or having these defaults?

Like that after the resolve:

```php
$this->readability = $config['readability'];
```

jtojnar: Good catch. We should probably switch that to typed class too.

jtojnar: Fixed by switching to typed class. Also did the same to `ConfigBuilderConfig`.

src/Extractor/ContentExtractor.php
```diff
@@ -126,18 +108,18 @@ public function process(string $html, UriInterface $url, ?SiteConfig $siteConfig
...
 
         // load and parse html
-        $parser = $siteConfig->parser();
-
-        if (!\in_array($parser, $this->config->getAllowedParsers(), true)) {
-            $this->logger->info('HTML parser {parser} not listed, using {default_parser} instead', ['parser' => $parser, 'default_parser' => $this->config->getDefaultParser()]);
-            $parser = $this->config->getDefaultParser();
+        $parser = Parser::tryFrom($siteConfig->parser());
+        if (null === $parser) {
+            $this->logger->info('HTML parser {parser} not listed, using {default_parser} instead', ['parser' => $siteConfig->parser(), 'default_parser' => $this->config->defaultParser]);
```
```suggestion
            $this->logger->info('HTML parser {parser} not listed, using {default_parser} instead', ['parser' => $siteConfig->parser(), 'default_parser' => $this->config->defaultParser->value]);
```

src/Graby.php
```diff
@@ -48,59 +50,20 @@ class Graby
...
-        $this->config = new GrabyConfig($config);
-
+    public function __construct(
+        private readonly GrabyConfig $config = new GrabyConfig(),
+        ?ClientInterface $client = null,
+        ?ConfigBuilder $configBuilder = null
+    ) {
         $this->logger = new NullLogger();
 
         // Debug mode can be activated with 'debug' => true
```
j0k3r:
```suggestion
        // Debug mode can be activated with GrabyConfig(debug: true)
```

README.md
```diff
@@ -211,68 +211,68 @@ configure it and inject it to `Graby\Graby`.
 This is the full documented configuration and also the default one.
 
 ```php
-$graby = new Graby([
+$graby = new Graby(new GrabyConfig(
```
j0k3r, line 224: Shouldn't we update the README to add `use` for all classes added?

j0k3r (approved):
Once again good job, if you want to squash some commits before merging, you can do it

PR #400, opened 2026-07-24
Normalize authors to avoid empty value

j0k3r:
Related: https://github.com/wallabag/wallabag/issues/8935

### DIFF
--- a/src/Extractor/ContentExtractor.php
+++ b/src/Extractor/ContentExtractor.php
@@ -187,7 +187,7 @@ class ContentExtractor
         $image = $extractedInfo['image'] ?? $image;
         $language = $extractedInfo['language'] ?? $language;
         $date = $extractedInfo['date'] ?? $date;
-        $authors = array_merge($authors, $extractedInfo['authors'] ?? []);
+        $authors = $this->normalizeAuthors(array_merge($authors, $extractedInfo['authors'] ?? []));
         $body = $extractedInfo['body'] ?? $body;
 
         // check if this is a native ad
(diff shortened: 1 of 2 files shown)


src/Extractor/ContentExtractor.php
```diff
@@ -682,6 +685,8 @@ public function process(string $html, UriInterface $url, ?SiteConfig $siteConfig
             );
         }
 
+        $authors = $this->normalizeAuthors($authors);
```
jtojnar, line 688: This is redundant. All previous assignments already pass through `normalizeAuthors`.

I guess this might be a defensive coding in case normalization is forgotten in one of the places?

src/Extractor/ContentExtractor.php
```diff
@@ -187,7 +187,7 @@ public function process(string $html, UriInterface $url, ?SiteConfig $siteConfig
         $image = $extractedInfo['image'] ?? $image;
         $language = $extractedInfo['language'] ?? $language;
         $date = $extractedInfo['date'] ?? $date;
-        $authors = array_merge($authors, $extractedInfo['authors'] ?? []);
+        $authors = $this->normalizeAuthors(array_merge($authors, $extractedInfo['authors'] ?? []));
```
jtojnar, line 190: From time complexity perspective, I do not like that we are normalizing already normalized authors again and again.

One way to fix that would be combining `normalizeAuthors` with `array_merge` and only normalize the second argument.

jtojnar:
Maybe introduce a helper like the following and then pass `$authors->values()` to `ExtractedContent`. This would both avoid the repeated normalization and prevent `$authors` containing non-normalized array being passed to `ExtractedContent` (since then, PHPStan would complain about missing `values()` method):

```php
/**
 * @internal
 */
final class NonEmptyStringSet {
    /** @var list<string> */
    private array $values = [];

    /**
     * @param list<string> $values
     */
    public function extend(array $values): void {
        foreach ($values as $value) {
            if (!\is_string($value)) {
                continue;
            }

            $value = trim($value);
            if ('' === $value || \in_array($value, $this->values, true)) {
                continue;
            }

            $this->values[] = $value;
        }
    }

    public function isEmpty(): bool {
        return $this->values !== [];
    }

    /**
     * @return list<string>
     */
    public function values(): array {
        return $this->values;
    }
}
```

PR #402, opened 2026-09-08
README: Analyze README examples with PHPStan

jtojnar:
For each PHP markdown code block in `README.md` not marked `no-extract`, we will extract the code into a separate file in `.extracted-code-examples` directory and run PHPStan on it. That should ensure the examples are at least somewhat valid and consistent with the codebase.

We could also run them with PHPUnit like guzzle-transcoder does but that would require transforming the source code to use mocking if we want any reliability at all.


README.md
```diff
@@ -75,50 +75,57 @@ $article = 'http://www.bbc.com/news/entertainment-arts-32547474';
...
+assert($result->getHtml() === "[unable to retrieve full-text content]");
+assert($result->getTitle() === "BBC - 404: Not Found");
+assert($result->getLanguage() === "en-GB");
+assert($result->getDate() === null);
+assert($result->getAuthors() === []);
+assert((string) $result->getEffectiveResponse()->getEffectiveUri() === "http://www.bbc.co.uk/404");
+assert($result->getImage() === null);
+assert($result->getSummary() === "[unable to retrieve full-text content]");
+assert($result->getIsNativeAd() === false);
+assert($result->getEffectiveResponse()->getResponse()->getHeaders() === […]);
```
j0k3r, line 128: I'm wondering how this could be valid for phpstan.

Maybe we should put a real comment instead

```suggestion
assert($result->getEffectiveResponse()->getResponse()->getHeaders() === [/* ... */]);
```

jtojnar: I believe it is parsed as a constant, no idea why does not complain that it is undefined. Did as suggested.

